PR #124, opened 2015-01-27
Change indentation so that it uses a variable you can set.

sirdude:
This could maybe use some work, but basically it sets INDENT_WIDTH to a define or a variable instead of hardcoding it to a specific value.


gdetrez:
The set of commits in here seems to do more than what you describe (changing the Makefile and parser among other things), is that intentional?

sirdude:
Sorry I haven't really done pull requests before, I think I maybe did it wrong.  I Would like to get all the changes I have made to my branch eventually into the main one.  You'll notice the first one: 557c720
Deals with the INDENT_WIDTH define  I committed that one to my branch and then made the pull request.  Then I continued to work on my branch, adding things.  I think each commit is pretty isolated but there are a few of them.  Whats the best way to do this, should I create a diff for each thing and then open an issue on it or is there a better way? :)

Thanks,

Kent

gdetrez:
I think the "github way" of doing this is to create a branch for each
issue/feature for which you want to create a pull request. If you're not
familiar with git, I think you should be able to do that using [git
cherry-pick](http://git-scm.com/docs/git-cherry-pick)

On Mon, Feb 16, 2015 at 4:31 PM, Kent Mein <notifications@github.com> wrote:

> Sorry I haven't really done pull requests before, I think I maybe did it
> wrong. I Would like to get all the changes I have made to my branch
> eventually into the main one. You'll notice the first one: 557c720
> <https://github.com/BNFC/bnfc/commit/557c720a8960d483bbe4a5297ab020588a594484>
> Deals with the INDENT_WIDTH define I committed that one to my branch and
> then made the pull request. Then I continued to work on my branch, adding
> things. I think each commit is pretty isolated but there are a few of them.
> Whats the best way to do this, should I create a diff for each thing and
> then open an issue on it or is there a better way? :)
>
> Thanks,
>
> Kent
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/BNFC/bnfc/pull/124#issuecomment-74525264>.
>



-- 
Grégoire Détrez

sirdude:
Ah, I'm going to hit close and comment on this one, if that doesn't kill it could you close this one.  I'll clean things up and submit 1 or two of them independently.

Thanks,

Kent

PR #129, opened 2015-02-07
Update README.md

DaveJarvis:
Requirements section.


gdetrez:
Thanks for this. I think it's useful to put that in the readme but I have a couple of comments:
- `apt` is only for debian and it's derivative (and to be a bit picky, debian is gnu/linux not unix :-))
- there is a couple more requirements (ghc, happy and some base library that are tied to the version of ghc) and I think the recomended way is to install the haskell-platform (`apt-get install haskell-platform`) which should contain everything needed to build BNFC.

PR #130, opened 2015-02-19
Use a define INDENT_WIDTH instead of hardcoding a number

sirdude:
for indentation on our pretty printers.  This makes it simpler for end
users to change it and I think makes the code a little more readable.


gdetrez:
Thanks!

PR #131, opened 2015-02-20
Add line number to yyerror for c

sirdude:
just like in the c++ implementation.
For c and c++ also added token that created the error to the output.


gdetrez:
Nice, thanks!

PR #132, opened 2015-03-09
Javascript backend using jison

olahol:
This is a backend for Javascript using Jison (https://zaach.github.io/jison/docs/). It's almost complete although there are some issues I have been unable to resolve:

1.  I have not gotten the tests under `testing/` to work, shelly gives me an error that compiled parsers such as `./Testgf` cannot be found.

2. The grammar for GF (`examples/GF/gf.cf`) does not work and I cant figure out why.


gdetrez:
This is great. Several people have asked for a javascript backend. :+1: 

> I have not gotten the tests under testing/ to work, shelly gives me an error that compiled parsers such as ./Testgf cannot be found.

I can help you with that but first the test script should read from stdin if there's no argument (many tests rely on that).

> The grammar for GF (examples/GF/gf.cf) does not work and I cant figure out why.

It seems that jison finds a lot of conflicts in `gf.cf` but there shouldn't be any. It's the same in C.cf (although there should be one shift/reduce there). I suspect a problem in the generation of the .jison file. I'll try to take a better look this week.

One more thing: Can you change the printer so that it output something more like the other backends and the book? So, with the Calc grammar it should be something like `EAdd (EInt 1) (EInt 2)` instead of `{"type":"EAdd","loc":{"start":{"line":1,...`.

olahol:
> I can help you with that but first the test script should read from stdin if there's no argument (many tests rely on that).

I will change the generation of Test*.js to fix this.

> It seems that jison finds a lot of conflicts in gf.cf but there shouldn't be any. It's the same in C.cf (although there should be one shift/reduce there). I suspect a problem in the generation of the .jison file. I'll try to take a better look this week.

That would be great, from what I can tell the .jison is almost identical to the .bison file.

> One more thing: Can you change the printer so that it output something more like the other backends and the book? So, with the Calc grammar it should be something like EAdd (EInt 1) (EInt 2) instead of {"type":"EAdd","loc":{"start":{"line":1,....

Sure, right now it just dumps the js object of the AST.

gdetrez:
I think I managed to track some reduce-reduce conflicts to a problem on how empty rules are handled.

With the following grammar:
```bison
%start BAR
%%
BAR : FOO "bar" | ;
FOO : | "foo" ;
```
jison reports a reduce/reduce conflict but there shouldn't be any (bison accepts it without conflicts). A workaround might be to use jison's lr mode (`jison -p lr`) but it seems to take a looooooooooooooooong time with bigger grammars like gf.cf.

I beleive this is related to this bug reported in the jison tracker: https://github.com/zaach/jison/issues/205

It seems that it doesn't create a conflict for every empty rule but only when two or more interact in certain ways (like in my example or the one from the issue above).

I'm not completely sure how to proceed right now. This bug is obviously a problem but your backend can definitely be useful with other grammars. On the other hand there might be other bugs in the generated GF parser (either jison bugs or bnfc bugs) but it might be difficult to track them as long as this one is there.

Maybe you can check that my example above is indeed a jison bug and report it to the jison developers. zaach/jison#205 can be mentioned as a possible duplicate but it might be something different (the interaction between the rules is not exactly the same). (Off course, if you feel like it you can try and fix it yourself and submit a pull request to zaach/jison :smile:)

On bnfc's side, we can probably still integrate your code and just skip the problematic grammars in the tests. But I'd like to add something to your todos: adding a short guide on how to use the jison dackend to the [documentation](https://bnfc.readthedocs.org/en/latest/user_guide.html) including a warning about the jison bug with a link to the bug report.

Does that sound ok to you?

olahol:
Wow that is some nice detective work :smile:

I have been looking at the code for jison and I don't think I am equal to the task of fixing this bug. We could of course report it and wait for it to be fixed or if you think the backend can be useful there is the option of marking it as incomplete or experimental in the cli.

Absolutely :+1:, I will write the documentation as soon as possible.

andreasabel:
Ping. (I am a new maintainer and am looking over the open PRs.)

What is the status of this PR?  Still interested? What is the progress with the TODOs?

andreasabel:
Happy new year, Ola!  And, did this PR make it onto the list of your new year resolutions? :-)

PR #133, opened 2015-03-11
Added OBJS define to Makefiles that are produced

sirdude:
This cleans things up a little on the Makefile and allows us to
make multiple binaries without linking *.o
(If your making a test program and a production program you don't
want to link the test.o to your final program.)

I also worked a little bit on making the files more similar so
it's easier to maintain them.  There is still more work to do here,
but this is a good start.  One thing to Note, I defined FLEX_OPTS and BISON_OPTS to make
 them similar but did not put the code in yet to actually use them in the files that didn't have them already.
I wanted to keep the pull manageable.

### DIFF
--- a/source/src/BNFC/Backend/C.hs
+++ b/source/src/BNFC/Backend/C.hs
@@ -61,28 +61,29 @@ makeC opts cf = do
 
 makefile :: String -> String -> String
 makefile name prefix =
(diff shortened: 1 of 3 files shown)


gdetrez:
Thanks. There is indeed a lot of duplicated code in the C/C++ backends that could probably be refactored into common modules.

PR #135, opened 2015-04-14
Make BNFC compile on GHC 7.10

mortberg:
The FlexibleContexts language extension and some type signatures when using elem are needed in order for BNFC to compile on GHC 7.10. For more information see:

https://ghc.haskell.org/trac/ghc/wiki/Migration/7.10


gdetrez:
Thanks! That should fix #137

PR #141, opened 2015-06-04
Adds usage function similar to haskell version to c and c++

sirdude:
Also cleaned up the code a little bit so it breaks lines at 80 chars...


gdetrez:
Hi. Sorry for the delay. Could you split this pull request? I'd definitely pull the usage function but I'm not convinced about splitting string literals.

sirdude:
Will do.  Sorry about that.

Kent

sirdude:
I removed the branch and started over to clean things up will recreate the pull request.

PR #142, opened 2015-06-10
Exposes writeFiles, makeHaskell and parseCF in the library

robstewart57:
The current way of using BNFC is as an executable. However, given that
BNFC is written in Haskell, why not expose its functionalities as
library calls to other Haskell programs?

This is the case of the cal-src project relating to the syntax of the
CAL actor language, which is a cabal-ised project that is a library
that exposes:

    library
      exposed-modules: AbsCAL
                     , LexCAL
         	     , ParCAL
                     , PrintCAL
                     , SkelCAL

The workflow in `Setup.hs` for cal-src is:

1. As a pre-build hook, run `writeFiles` with `makeHaskell` to create
   files AbsCAL.hs, LexCAL.hs, ParCAL.hs, PrintCAL.hs and SkelCAL.hs
   corresponding to the "CAL.cf" grammar file in the cal-src
   repository.

2. Run cabal to create the cal-src library.

The `Setup.hs` for cal-src is:

    main :: IO ()
    main = defaultMainWithHooks simpleUserHooks { preBuild = calGenHook }

    calGenHook :: Args -> BuildFlags -> IO HookedBuildInfo
    calGenHook _ _ = do
      let opts = defaultOptions { lang = "CAL" }
      grammar <- readFile "CAL.cf"
      cf <- parseCF opts TargetHaskell grammar
      writeFiles "." $ makeHaskell opts cf
      return emptyHookedBuildInfo

The move away from calling `bnfc` within Haskell to using the
(proposed) exposed BNFC library is here:

https://github.com/robstewart57/cal-src/commit/c9165159488c19de7393d72c07fd55bd4e4dab31


robstewart57:
Relates to this mailing list discussion:
https://groups.google.com/forum/#!msg/bnfc-dev/9_37G02ErDw/yIa2biUoOXgJ

gdetrez:
Hi. I saw the discussion and I understand your use case.
I'm not really comfortable making so much internals of BNFC public as I don't want to commit to keeping those interface stable. What I could imagine is to have a separate module, say Distribution.Bnfc that would expose a stable API just for cabal usage.

But maybe there is some advantages in calling the bnfc executable instead of taping into the library. In particular I think it would be consistent with the way other tools (happy, alex or even ght) are used. One might for instance expect to use the --with-bnfc=... or --bnfc-option=... cabal options.

robstewart57:
Hi @gdetrez ,

Perhaps a third option, closely related to your 2nd option, is to patch cabal to have knowledge of `bnfc` as an executable, to be used in the `build-tools` field of a Haskell project. It relates to this SO answer:
http://stackoverflow.com/a/24744052/1526266

> "Alternatively, add Compiler.RSL.Syntax.Lex to your 'exposed-modules' list. This tells cabal that the Lex.hs file should exist, and so if it doesn't cabal will look for ways to build it using the tools in the build-tools line."

That is, to add bnfc to:

    builtinPrograms :: [Program]

In: https://github.com/haskell/cabal/blob/master/Cabal/Distribution/Simple/Program/Builtin.hs

gdetrez:
That would indeed be very cool :+1:

PR #145, opened 2015-06-24
A little bit of code clean up.

sirdude:
This mostly removes blank lines trying to standardize the
code a bit.  It also changes a couple of lines so they are more readable
on a terminal.


gdetrez:
I'm so sorry. Most of the files you edited are auto-generated by bnfc itself so the changes will be overwritten every time we re-generate them (like I did this morning :confused:). If you want to improve those files, you need to change the backend that generates them (src/BNFC/Backend/Haskell).

If you want, you can open a new PR with the other files though (Main.hs and BNF.cf)

PR #148, opened 2015-07-01
Improve the generated code for c and c++

sirdude:
Just makes the syntax more consistant and hopefully more readable.

### DIFF
--- a/source/src/BNFC/Backend/C/CFtoCPrinter.hs
+++ b/source/src/BNFC/Backend/C/CFtoCPrinter.hs
@@ -73,11 +73,11 @@ mkHFile cf groups = unlines
  ]
  where
   eps = allEntryPoints cf
-  prPrints s | normCat s == s = "char* print" ++ s' ++ "(" ++ s' ++ " p);\n"
+  prPrints s | normCat s == s = "char *print" ++ s' ++ "(" ++ s' ++ " p);\n"
     where
       s' = identCat s
   prPrints _ = ""
-  prShows s | normCat s == s = "char* show" ++ s' ++ "(" ++ s' ++ " p);\n"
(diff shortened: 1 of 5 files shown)


gdetrez:
Hey @sirdude, can you update the broken doctests to match your changes?

sirdude:
Sure no problem. This is done...
I'm having a little problem.  I'm not sure how to run the tests on my own machine.  I've been building with cd source; make; make install
I can build the doctest with
ghc test/doctests.hs
./test/doctests


When I try to do that with the unit-tests.hs I get this:
ghc test/unit-tests.hs
test/unit-tests.hs:18:18:
    Could not find module `BNFC.WarningMSpec'
    Use -v to see a list of the files searched for.

gdetrez:
Thanks! Looks good now :smile: 

I think the problem is that you need to tell ghc where to find the sources: `ghc -isrc/ -itest/ ... test/unit-tests.hs`. The easiest way to run the tests is to use cabal, the following should build and run the tests in one go:
````
cabal configure --enable-tests && cabal test
```

You might need to install the test libraries:
``` 
cabal sandbox init && cabal install --only-dependencies --enable-tests
```

sirdude:
Ok I think this is the issue, inside of src/bnfc/source
cabal configure --enable-tests
        works
cabal build
        gives this:
        Preprocessing test suite 'unit-tests' for BNFC-2.8...

src/BNFC/CF.hs:118:8:
    Could not find module `LexBNF'
    Use -v to see a list of the files searched for.


cabal test gives this:
Running 2 test suites...
cabal: Error: Could not find test program
"dist/build/bnfc-doctests/bnfc-doctests". Did you build the package first?

Sorry I'm a haskell newbie.  I think it's not building everything though so unable to run the tests.  I'm guessing I need to add a path like you mentioned before but not sure how to do that with cabal.

Thanks,

Kent

PR #155, opened 2015-10-23
Add ANTLR4 Java backend 

gapag:
I wrote an [ANTLR4] (http://www.antlr.org/) backend hopefully substituting the old J(F)Lex+CUP . 
ANTLR is a lexer+parser generator originally targetting Java. It has many commonalities with BNFC, but it deals mainly with parse trees whereas BNFC has its focus on ASTs.

The motivation is that 
 1. these tools are old and at least one of them is not supported anymore.
 2. ANTLR has interesting features to manipulate parse trees, which could be integrated in BNFC.
 3. ANTLR has also some other target languages (e.g. Python)

Commit c7ba720 's message contains further details about what has been done and what is to do. 

Summarizing, what I did is
- make ANTLR a non-default option for the Java backend
- CF to ANTLR lexer/parser translation
- Java Makefiles calling ANTLR
- Blacklist Java keywords (abstract, native ...) (ONLY for ANTLR backend)

System tests pass successfully on my machine.

The directory .idea is created by the IDE I use and you can of course remove it from the merge -- unless there are other people willing to use my same project.

I did not set ANTLR as the default Java backend because I'd like to have more people trying it before promoting it.

Thanks for your time.


gapag:
Oops. Doctests fail. I will fix this on Monday. I will also fix the classpath in the example tests, right now there is no antlr library in the javatools.jar. Sorry!

.idea/libraries/org_antlr_antlr4_4_5_1_1.xml
```diff
@@ -0,0 +1,12 @@
+<component name="libraryTable">
```
gdetrez: This seems a bit too IDE-specific to be in the repository.
Although, I like the idea of using some package manager for java (maven?) to get the dependencies instead of the (rather ugly) javatools.jar...

README.md
```diff
@@ -54,7 +54,7 @@ https://bnfc.readthedocs.org/en/latest/
 Run tests
 ---------
 
-Here is how to build and run the different test suites for bnfc.
+    Here is how to build and run the different test suites for bnfc.
```
gdetrez: Typo?

source/src/BNFC/Backend/Java.hs
```diff
@@ -97,138 +121,495 @@ makeJava options@Options{..} cf =
...
+    fileextension::String,
+    toolname::String,
+    toolversion::String,
+    supportsEntryPoints::Bool,
+    results :: [String],
+    moveresults :: Bool
+}
+
+-- Positional constructor.
+mkMakeFileDetails :: String -> (OutputDirectory->String) -> String -> String -> String -> String -> Bool -> [String] -> Bool -> MakeFileDetails
```
gdetrez: This looks like an alias for `MakeDetails`, is it really necessary?

.idea/libraries/org_antlr_antlr4_4_5_1_1.xml
gapag: You can remove it, by itself it does not help to automate the dependencies -- it is not a maven input file AFAIK. 

I write the following just so that you can follow my line of though and see if it is the same of yours:
I guess it should be necessary to have some maven file with such dependencies and run maven if necessary when you run the tests, which implies you should have maven on your test server.

README.md
gapag: Yes.

source/src/BNFC/Backend/Java.hs
```diff
@@ -97,138 +121,495 @@ makeJava options@Options{..} cf =
...
+    ,"class BNFCErrorListener implements ANTLRErrorListener"
+    ,codeblock 2 [ "@Override"
+        ,"public void syntaxError(Recognizer<?, ?> recognizer, Object o, int i, int i1, String s, RecognitionException e)"
+        ,codeblock 2[ "throw new"<+>tedoc<>"(s,i,i1);"]
+    , "@Override"
+        ,"public void reportAmbiguity(Parser parser, DFA dfa, int i, int i1, boolean b, BitSet bitSet, ATNConfigSet atnConfigSet)"
+        ,codeblock 2[ "throw new"<+>tedoc<>"(\"Ambiguity at\",i,i1);" ]
+    ,"@Override"
+        ,"public void reportAttemptingFullContext(Parser parser, DFA dfa, int i, int i1, BitSet bitSet, ATNConfigSet atnConfigSet)"
+        ,codeblock 2 [{-"throw new"<+>tedoc<>"(\"Attempting full context\",i,i1);"-}]
```
gdetrez: This looks suspicious, left-over debug code?

source/src/BNFC/Backend/Java.hs
```diff
@@ -97,138 +121,495 @@ makeJava options@Options{..} cf =
...
+    ,"@Override"
+        ,"public void reportContextSensitivity(Parser parser, DFA dfa, int i, int i1, int i2, ATNConfigSet atnConfigSet)"
+        ,codeblock 2 [ {-"throw new"<+>tedoc<>"(\"Context sensitivity\",i,i1);"-}]
```
gdetrez: same here

source/src/BNFC/Backend/Java.hs
gapag: I am not sure! Being not very expert at Haskell I essentially wanted this function to have a curryfiable constructor.
This is repeated for some other record types, I guess.
I kind of made me more familiar with the record syntax and behaviour so I think I can review the code to remove this usage, which I suspect defeats the advantages of using records.
Do you want me to do this?

.idea/libraries/org_antlr_antlr4_4_5_1_1.xml
gapag: I don't have direct experience using Maven anyway, I usually use it through IDEs.

gdetrez: Yes, that's what I was thinking. Maven is usually available as a distribution package so it shouldn't be a problem to have it on the test server.

source/src/BNFC/Backend/Java.hs
gdetrez: You can always use a record constructor as a positional constructor, e.g.:
```haskell
λ data Foo = Foo { foo :: String } deriving Show
λ Foo {foo="bar"}
Foo {foo = "bar"}
λ Foo "bar"
Foo {foo = "bar"}
```
So you can just delete the function and replace every instance of `mkMakeFileDetails` by `MakeDetails`

source/src/BNFC/Backend/Java.hs
gapag: No, essentially is the machinery needed to have ANTLR behave like CUP. If ANTLR cannot parse something, it does not throw any exception and goes on.
The reason this boilerplate code is to simulate the behaviour of the CUP backend, so that tests can expect the same outcome from the ANTLR backend.
Should I prepend a Java comment to such additional generated code explaining that it is there to launch an exception whenever the parser cannot parse something?

source/src/BNFC/Backend/Java.hs
gapag: Thank you. I will take care of it and push a new commit.

source/src/BNFC/Backend/Java/CFtoJLex15.hs
```diff
@@ -45,6 +45,13 @@ import BNFC.Utils (cstring)
 import BNFC.Backend.Common.NamedVariables
 import Text.PrettyPrint
 
+-- This is for now the workaround to differentiate JLex and JFlex.
```
gdetrez: I see the point about readability but given the small differences between the two versions, I don't think having two different functions would be any better... You could pass the `JavaLexerParser` value if you prefer (although that means raising an error on `Antlr4`)

Or maybe create a datatype that encapsulate the difference between the two variants? (something like `data JLexVariant = JLexVariant { bufferVar :: String, escapeChars :: [Char], ... }`)

.idea/libraries/org_antlr_antlr4_4_5_1_1.xml
gdetrez: No problem, I was just thinking out loud :-)
Can you remove the .idea directory though?

source/src/BNFC/Backend/Java.hs
gapag: Mm now I understand why I did so for some types. It was for sparing me to write the name of the field to initialize.
Nevertheless I will follow your advice, the code will anyway be more readable and less prone to have bugs introduced.
it is hard to say which is which when you use the positional constructors.

source/src/BNFC/Backend/Java.hs
gdetrez: My comment was only concerning the last line, where you have a commented string restulting in an empty `codeblock`

source/src/BNFC/Backend/Java/CFtoJLex15.hs
gapag: This is surely a comment I should have removed. 
I don't think it is *terribly* bad. I get nuisance from see the checks spreaded and having boolean arguments to functions.
I agree completely with your observation about having two different functions.
For your proposal I think it's OK leaving the code as it is, it is probably not worth the effort.
I beg your pardon for using such an inappropriate hyperbole.

.idea/libraries/org_antlr_antlr4_4_5_1_1.xml
gapag: Yes, I will make some commits to address all your comments.

gapag: Before tonight.

source/src/BNFC/Backend/Java.hs
gapag: Ah ok. Yes, I will remove them. No exception should be thrown when these handlers with commented code are executed. They execute in response to events occurring in ANTLR when predicting which will be the next entity, and are of no interest for the tests.

gdetrez:
Hi Gabriele.

Sorry it took me so long to get back to you on this. I think this is definitely a great addition to bnfc as we have sort of fallen behind with respect to the parsing tools available for java.

Globally, this looks good and coherent with the rest of the code. (Thanks in particular for taking the time to add system tests!!). So most of my comments at this point are about polishing.

Can you look at the following in addition the the line comments:

* I don't like `TODO`s and `FIXME`s. First they have a tendency to stick around forever and second they make it difficult to have a discussion about the problem. I suggest the following:
  * if it's trivial, just fix it
  * if it's more work but it doesn't really matter (like potential refactorings), just remove the TODO
  * for anything else (future work, questions, potential bug...), open an issue
* There is a few places where the code could be simplified a bit (e.g.: `( "PARSER", (executable parmake))` -> `( "PARSER", executable parmake),`, you can run hlint at least on the new files to get reformating suggestion (you can run hlint on the file you only modified but you'll get a lot of suggestion about the old code as well... :-/)

As a FYI, I'm globally trying to follow [this](https://github.com/tibbe/haskell-style-guide/blob/master/haskell-style.md) as a style guide, although like for hlint, the legacy code is still a mess at this point...

gapag:
Great, I will try to stick to the style guide and polish the code as much as I can. I will push the commits before tonight. Thanks to you for reviewing the code.

gapag:
I meant haddock not HSpec. Shame on me.

source/src/BNFC/Backend/Java/CFtoJLex15.hs
```diff
@@ -45,6 +45,10 @@ import BNFC.Utils (cstring)
 import BNFC.Backend.Common.NamedVariables
 import Text.PrettyPrint
 
+
+cf2jlex' str cf = cf2jlex str cf False
+cf2jflex' str cf = cf2jlex str cf True
+
```
gdetrez: I still think this needs fixing. I suggest the following solution: reorder the argument of `cf2jlex` and use `cf2jlex True` and `cf2jlex False` instead.

Or, if you really want two different functions that's ok but give them names without quotes.

gapag: I'll follow your advice, I'll fix this before next week.

gdetrez:
This is great!

One last thing, if I may: can you write a short help section about the antlr backend in the doc? It doesn't have to be very long, just a quick example of how you use it is enough. You can look at [what I wrote for pygment](http://bnfc.readthedocs.org/en/latest/user_guide.html#pygments) for an example.

gapag:
I will do it, before next week too.

gdetrez:
Oh, I somehow missed the notification that you updated this branch :-/
Thanks for your fixes, merging now!

PR #161, opened 2015-11-28
Compile with GHC 7.10.2, Alex 3.14

johnjcamilleri:
I added these lines to get BNFC to work with GHC 7.10.2, Alex 3.14.
I guess these additions won't work with earlier versions, so some version-checking is still required in the code.


johnjcamilleri:
This probably won't work with older versions of GHC/Alex, so some version checking probably needs to be implemented.

gdetrez:
Thanks!

danr:
Can you push this as a new version on Hackage?

danr:
bumpedy-bump : )

gdetrez:
I just uploaded 2.8.1 on hackage that should fix this problem. Could you try it and confirm that it works (I wanted to make it a package candidate but I didn't find the checkbox for that :-/)

Sorry it took so long...

PR #165, opened 2016-02-03
FlexibleContexts added to doctest arguments

pascalh:
Hi, I had problems running the test suite with ghc-7.10.2, because the extension FlexibleContexts was missing in the arguments to doctest. 

Running the tests as described in the readme lead to a compilation error, since one has to run `cabal build` before actually running the tests.


gdetrez:
Thanks!

PR #166, opened 2016-02-05
The precedence of reserved symbols and predefined tokens was different in Haskell and OCaml versions.

tetsuo-jp:
-- I fixed the order in the OCaml backend.
-- The following file produces different results in Haskell and OCaml
versions.
{-
% bnfc -m -ocaml A.cf
% make
% echo ":a" | ./TestA
[Abstract syntax]

MyId (MyIdent ":a")

[Linearized tree]

:a
% bnfc -m -haskell A.cf
% make
% echo ":a" | ./TestA

Parse Successful!

[Abstract Syntax]

Keyword

[Linearized tree]

:a
-}

Keyword. Exp ::= ":a" ;
MyId.    Exp ::= MyIdent ;

token MyIdent (':' letter)


gdetrez:
Thanks :+1:

PR #173, opened 2016-03-07
create cabal files

pascalh:
This pull request allows bnfc to create cabal files in order to build/install the generated modules. 

(I will start working on a cabal hook, which lets you use bnfc as a preprocessor to cabal.)

### DIFF
--- a/source/BNFC.cabal
+++ b/source/BNFC.cabal
@@ -50,7 +50,8 @@ Executable bnfc
     containers,
     pretty >=1.1 && <1.2,
     filepath,
-    deepseq
+    deepseq,
+    Cabal == 1.22.*
   if  impl(ghc < 7.6.1)
     Build-depends: ghc-prim
   build-tools:         alex, happy
@@ -94,6 +95,7 @@ Executable bnfc
     BNFC.Backend.Haskell,
     BNFC.Backend.Haskell.ToCNF,
     BNFC.Backend.Haskell.RegToAlex,
+    BNFC.Backend.Haskell.CFtoCabal,
     BNFC.Backend.Haskell.CFtoTemplate,
     BNFC.Backend.Haskell.CFtoAlex3,
     BNFC.Backend.Haskell.CFtoAlex2,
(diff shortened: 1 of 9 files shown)


gdetrez:
Hi @pascalh. Thanks for this :+1: and for taking time to add tests :heart:

I see from the code that it should add the proper dependencies but when trying on my local machine, the generated `build-depends:` is empty, any idea why that might be?

Also, would it be possible to also build the test program?

pascalh:
I can not imagine why the dependencies are not being added on your machine. Which ghc and cabal version are you using? If this is a major problem we could generate the cabal files without cabal library (using plain text concatenation).

Building the test program is not straight forward, because the module and file name differ. I could rename the module Main and everything works fine. I felt like this is a bigger change, which should not be implemented without any discussion.

gdetrez:
It took me a while but I manage to make some more tests. I am using ghc 7.8 and Cabal (library) 1.18. I tried rebuilding bnfc with Cabal 1.22 and it now correctly populates the list of dependencies. So it might just be a bug in old Cabal library and I guess one solution is to make the lower version bound higher (1.22 seems to work, 1.20 might as well)

I'm not sure I understand what is the problem with the test executable. I tried adding a section to the cabal file manually and it seems to work fine.
What I added looks like this (there might be a way to make it less redundant though...)
```cabal
Executable TestCalc
   build-depends:
     base >4,
     array -any
   main-is: TestCalc.hs
   Other-modules:
     AbsCalc
     ErrM
     PrintCalc
     LexCalc
     ParCalc
```

pascalh:
Thanks for testing. It worked out well using your approach.

andreasabel:
Seems like @gdetrez has moved on, this threat has been dormant for a year.
I am the new guy trying to take over some maintenance for bnfc (I use it for my class).

What is the status of this feature @pascalh ?
- [ ] Still interested.
- [ ] Good test coverage.
- [ ] Good documentation (integration into bnfc-tutorial, for instance).

pascalh:
@andreasabel Great to see BNFC coming back to life. I will update the pull request as soon as I find some time to do so.

andreasabel:
Possibly `cabal init` could be utilized to generate the cabal file, see https://github.com/haskell/cabal/pull/7253 .
I also played with `hpack`, see https://github.com/sol/hpack/issues/406 .

Issue #172, opened 2016-03-04
[Bug] Bison file prefixes for C/C++ not generating correctly


dcavar:
When generating C or C++ and using prefixes, the generated Bison-file is missing the prefixes in the error function. Example (take any well-formed and otherwise working *Test.cf*):

`bnfc -m --cpp -p test Test.cf`

the generated *Test.y* will have missing prefixes in the error function:

```
void testyyerror(const char *str)
{
  extern char *testyytext;
  fprintf(stderr,"error: line %d: %s at %s\n", 
    yy_mylinenumber, str, yytext);
}
```

but it should prepend the prefixes on all the variable references in that function:

```
void testyyerror(const char *str)
{
  extern char *testyytext;
  fprintf(stderr,"error: line %d: %s at %s\n", 
    testyy_mylinenumber, str, testyytext);
}

```

pascalh:
I applied the suggested changes to the C++-backend in my [branch](https://github.com/pascalh/bnfc/tree/issue172). Please confirm that my changes solve the issue.

dcavar:
Dear Pascal,

thanks a lot! I just tested the new code, and it seems to work for
single character delimited comments using "some comment".

Thanks again!

DC


On 03/08/2016 08:20 AM, Pascal Hof wrote:
> I applied the suggested changes to the C++-backend in my branch
> <https://github.com/pascalh/bnfc/tree/issue172>. Please confirm that my
> changes solve the issue.
> 
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/BNFC/bnfc/issues/172#issuecomment-193782642>.
> 

-- 
Damir Cavar
https://linguistlist.org/people/damir_cavar.html

gdetrez:
I believe that @pascalh's pull request #174  is solving this issue. @dcavar if you have time to test it and there is still a problem, feel free to re-open it.

PR #174, opened 2016-03-08
Generating prefixes in CPP (issue 172)

pascalh:
These changes should solve issue #172 and thus make the cpp-backend generate code with prefixed names correctly.


gdetrez:
Thanks!

PR #177, opened 2016-04-22
Python

gapag:
Hi, I would like to contribute with a python backend. It is implemented with Antlr4. I did not test it thoroughly, I could not run the system tests (I am becoming more and more rusty with computers) and I will try it again when I am less tired.

I need still to run hlint and to make the code follow the guidelines I was once pointed to, so no hurry.


gapag:
Ah, now I recalled how to run them.

gdetrez:
Hi @gapag. If I recall correctly, you told me that this was not working as well as you wanted to, is that right? Do you still want it merged or should I close the PR until you have time to work more on it?

gapag:
I can close it.
Antlr4's python output did not work properly and it is not a BNFC problem.

gdetrez:
Ok, thanks. Maybe you can keep the branch around in case someone else (or maybe yourself ;-)) wants to give an other shot a python with a different parser. I'm guessing they could re-use at least the AS and pretty printer modules.

gapag:
I think ANTLR4 can be replaced with Grako, a python library.

Grako generates model classes and visitor classes, so it would simply be 
a matter of translating the grammars and reuse the prettyprinter (and 
throw away the rest :)

Let's see if I find some time to work on it before Christmas. I think it 
might take simply a weekend to have something working.


On 2016-08-23 10:05, Grégoire Détrez wrote:
> Ok, thanks. Maybe you can keep the branch around in case someone else
> (or maybe yourself ;-)) wants to give an other shot a python with a
> different parser. I'm guessing they could re-use at least the AS and
> pretty printer modules.

---
Detta e-postmeddelande har sökts igenom efter virus med antivirusprogram från Avast.
https://www.avast.com/antivirus

PR #185, opened 2016-08-26
[C] Emit correct function prototypes

BenWiederhake:
This affects the generated file `Absyn.h`.

Example before:
```
    Full_combinator_id make_AbsentFCID();
```
Example after:
```
    Full_combinator_id make_AbsentFCID(void);
```

This change is required as `make_AbsentFCID()` in C actually
means "takes an arbitrary amount of arguments", which is not
the desired semantics.

`make test` passes all suites.  Someone should probably write a regression test for this, so this bug isn't introduced again.


gdetrez:
Thansk!

PR #187, opened 2016-10-24
Haskell-backend: printer: better heuristics for space after closing paren

andreasabel:
Instead of

```
  f (g x)y : (a -> b)-> c
```

prints now

```
  f (g x) y : (a -> b) -> c
```

I have not looked at the other backends yet.  Are the generated printers kept in sync?  Are there test-cases to verify the output of the generated printers on some examples?


gdetrez:
@andreasabel Sorry, I missed this PR before for some reason...

> Are the generated printers kept in sync? Are there test-cases to verify the output of the generated printers on some examples?

I think they should be more or less in sync but no, there is no automated test of the pretty-printers' output. We could relatively easily check the output of the test program though...

andreasabel:
Thanks!

Issue #186, opened 2016-10-12
Detect and report invalid identifiers


gdetrez:
_This issue was first discussed in the mailing list_

The [lbnf reference](http://bnfc.readthedocs.io/en/latest/lbnf.html#basic-lbnf) specify that "An identifier is a nonempty sequence of letters, starting with a capital letter." but bnfc will accept a rule such as `cat . Type = "Foo";`, generating invalid haskell code.

Instead, bnfc should detect the invalid identifier and print a error message.

pascalh:
I'd like to work on that issue as I already said on the mailing list (if it is fine for you).

gdetrez:
@pascalh yes, I saw, thanks for volunteering!
 It seems that I can only assign issue to member of the organization. I sent you an invitation :-)

andreasabel:
It seems that lower case identifiers are for __defined__ identifiers, via the `define` pragma.  This feature is documented in the book _Implementing Programming Languages_, see http://www.grammaticalframework.org/ipl-book/ , but not in the user manual.
Also, there is no check that a definition is actually supplied in the grammar file.  A badly implemented and badly documented feature which eventually got broken by some refactoring, see #214.

andreasabel:
I did serious work on the `define` feature in #214, #266, and #285, addressing the lowercase label issue. Error reporting is better now.  Also, apostrophes are now forbidden in identifiers, see #279.

Closing this now, feel free to reopen with new information.

PR #188, opened 2016-10-25
refuse rule labels starting with lower case letter

pascalh:
this pull request is solving issue #186.

### DIFF
--- a/docs/othertools.rst
+++ b/docs/othertools.rst
@@ -1,6 +1,17 @@
 =============
-Similar tools
+Other tools
 =============
+
+LBNF tools
+----------
+`LBNF for Sublime Text`_
+  Provides syntax highlighting for LBNF in Sublime Text.
+
+.. _LBNF for Sublime Text: https://github.com/Centril/sublime-lbnf-syntax
+
+
+Similar tools
+-------------
 
 bnfc-meta_
     Like BNFC for haskell but allows you to define your grammar in a .hs file
@@ -20,4 +31,3 @@ ANTLR_
 .. _ANTLR: http://www.antlr.org/
 .. _BNF for Java: http://bnf-for-java.sourceforge.net/
 .. _GF: http://www.grammaticalframework.org/
-
--- a/source/changelog
+++ b/source/changelog
@@ -1,3 +1,22 @@
+2.9.0 (Development version)
+	* C: Emit correct function prototypes
+	* Java: Add an experimental ANTLR4 backend
+	* C, C++, OCaml, Java: Accept ' or " as comment delimiters [#146]
+	* Improve error messages [#144]
+	* Haskell: Generated code mostly hlint-warning free
+	* bnfc is now more accepting about border-line terminals [#149]
+	* Cleaner error message on undefined categories [#151]
+	* C++: Fix regression introduced in 2.8 in Squeleton.H [#164]
+	* C++: Replace `%name-prefix` with `%define api.prefix` in bison files
+	* C++: Fix a bug that appeared if you had a category named "List"
+	* C, C++: Add usage output to the test program
+	* C, C++: Fix a bug in the parser file when using -p [#172]
+	* OCaml: Reserved symbols now have a higher precedence than predefined
+	  tokens as in Haskell
+
+2.8.1 Grégoire Détrez <gregoire.detrez@gu.se> May 2015
+	* Fix compatibility with GHC 7.10.2 and Alex 3.14
+
 2.8 Grégoire Détrez <gregoire.detrez@gu.se>  May 2015
 	* Builds with ghc 7.10.1
 	* Add support for JFlex (java)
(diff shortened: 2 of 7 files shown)


gdetrez:
Thanks @pascalh. `ParBNF` is actually generatef from `BNF.cf` by bnfc, so it's not safe to put custom code there. I think the best place for this would be in `BNFC.GetCF`. It's a bit of a mess and I think the best way is to add a new step in the sort of processing pipeline line [48-51](https://github.com/BNFC/bnfc/blob/master/source/src/BNFC/GetCF.hs#L48).

pascalh:
Thanks for pointing that out, @gdetrez. The module `GetCF` already contained the rule label check in a comment.

gdetrez:
That's... interesting :confused: 

I was wondering if there was a reason for that, but sadly the repository history doesn't go that far.
That makes me a bit wary of just un-commenting it...

What about we make it a warning for now, instead of exiting? The message could include something like "This will become an error in a future version".

pascalh:
Going even further bnfc could capitalize respective rule labels in order to ensure correct compilation of the generated sources. My point is that bnfc can identify the problem and solve it without confusing the user. I hereby assume that automatic capitalization is not confusing to someone using a compiler-generator or even a compiler-generator-generator. ;)  Anyway I can understand your worries about the un-commenting, so a warning could also do it if you do not like my proposal.

The other issue that bothers me about the function [checkCode](https://github.com/BNFC/bnfc/blob/master/source/src/BNFC/GetCF.hs#L382) is that the error messages are not of much help for the user, because they just say "Bad xyz". At least for the manually checked properties of a grammar we could offer more helpful error messages. This seems like a new issue to me.

gdetrez:
I agree that it would be better to automatically handle capitalization. It is done in some places, but probably not everywhere.
There is a similar issue with underscores. The manual says only letters and numbers but I think bnfc won't complain if you have one in a label/category name and leave it to the language to accept the name or not. I don't know if any of the target language would reject a name with an underscore but they are more idiomatic in some languages than others.

I also agree that the error message leave room for improvement. I'm not sure if you're you offering to work on that or suggesting that we open a different issue about it?

pascalh:
Having thought about this issue again. Consider the following (false) grammar:
`f . Foobar ::=  "foobar" 
 F . Foobar ::= "Foobar"`

The user expects that a type with two different constructors is being created. In this case I would prefer a bnfc (!) warning or an error because of a lower case rule. An automatic capitalization in this case creates false Haskell code or yields a (to the outside world) strange bnfc error, because bnfc does something one would not expect. 

This is why I changed my mind and suggest sticking to the manual and refusing every false input, such that the user has to take care himself about possible problems in the definition of the grammar. 

Another option could be an automatic correction, while also printing a warning. Somehow I lost my confidence in the automatic corrections.

The improvements of the error messages should just be a new issue. I might work on that someday, but I prefer working on other stuff before.

gdetrez:
Good point. Then let's do the progressive thing where we print a warning now that will be turned into an error later.

Since the  the lbnf specification says:

> An identifier is a nonempty sequence of letters, starting with a capital letter.

So should we also ban `_` and numbers?

pascalh:
I like the progressive way starting with a warning for lower case rules labels.

Currently I feel way safer also warning the user not to use underscores and numbers, because they violate the spec and could potentially lead to problems - even if I cannot think of such a use case. Do you know why these characters are banned for identifiers?

gdetrez:
No I don't really know but I have some guesses... The numbers are used for coertions and although that's only at the end of the token, it might have been easier to ban them completely instead of having a more complex rule.

For underscores I'm guessing the limitation comes from bnfc itself. For instance if you have a label `Function` that happens to be a keyword in your target language, I think bnfc might use `function_` instead. It also allows to create more idiomatic function names. For instance, turning a `FooBar` into `foo_bar` in a language where snake case is the preferred style without risking a name conflict.

pascalh:
@gdetrez the pull request is now ready for merge.

bafain:
Running this on the semantic definitions [example](https://github.com/BNFC/bnfc/blob/35df5ce/examples/DefinedRules.cf) raises the following warnings:
```
Warning: bad constructor name if in if. Stm ::= "if" "(" Exp ")" Stm "endif"
Warning: bad constructor name for in for. Stm ::= "for" "(" Stm ";" Exp ";" Stm ")" Stm
Warning: bad constructor name inc in inc. Stm ::= Ident "++"
Warning: bad constructor name op in op. Exp ::= Exp1 Op1 Exp1
Warning: bad constructor name op in op. Exp1 ::= Exp1 Op2 Exp2
```
Is this intended? Note that the lowercase labels refer to the following function definitions:
```
define if e s	   = If e s (Block []) ;
define for i c s b = Block [i, While c (Block [b, s])] ;
define inc x	   = Assign x (EOp (EVar x) Plus (EInt 1)) ;
define op e1 o e2  = EOp e1 o e2 ;
```
While semantic definitions seem to be broken in BNFC (and already were at the time of #186 and #188), something like the following adaption of the example in https://github.com/BNFC/bnfc/issues/186#issue-182529854 might have been a valid use of lowercase labels once:
```
cat. Type ::= "Foo" ;
internal Cat. Type ::= ;
define cat = Cat ;
```

andreasabel:
@bafain Indeed, lower case identifiers are to be used with `define`, but BNFC does not throw an error if there is no `define` matching the use of an identifier.

andreasabel:
After the fix of #266, the `define` facility is available for all backends.
Thus, I close this PR, since lower case letters do have a meaning in BNFC.

Issue #198, opened 2016-11-24
Use Stack to build bnfc


pascalh:
Recently [stack](https://haskellstack.org/) gets more and more popular, since it avoids cabal-hell, eases updating used packages and ghc and also uses a sandbox for every project by default. Stack also allows the ease use of different versions of ghc on a system. Hereby I suggest to use stack as a build tool for bnfc.

metaleap:
Did this PR get accepted? If not, and anyone comes across this: one can, in the `sources` sub-directory, do a `stack init` (creates a stack.yaml from the .cabal file) then `stack install`, this worked at my end. Only oddity is that using the current git repo you get 2.8 whereas if you don't clone at all, ever and just do `stack install BNFC` *outright* instead, you already get 2.8.1 😏

pascalh:
@metaleap waiting for an answer regarding this PR. This is exactly what I did to raise this PR ;)

Master is the dev branch and you find the branch for 2.8.1 [here](https://github.com/BNFC/bnfc/tree/2.8.x). However, the version number in master should be bumped to 2.8.1 as well.

ip1981:
First of all you can use Stack as is. If you find yourself in need of many files and configuration for Stack you are in Stack-hell.

Second, cabal-hell is no longer a problem since we have new builds with http://cabal.readthedocs.io/en/latest/nix-local-build-overview.html

And `cabal new-freeze` is a way better than Stack's snapshots, because information is local. But both mean flaws in design and Haskell code quality. Good old `.cabal` file ought to be enough with minimal version requirements and packages must maintain backward compatibility.

P. S. Yes, I dislike Stack.

rabbitonweb:
@ip1981 so.... instead of helping with the issue, you've decided to spent this time to tell us how you don't like Stack.. well, that was soooooo helpful! Thank you! It moved this issue forward.

P.S Yes, irony
P.P.S Being ridiculous is ridiculous

andreasabel:
#199 has been merged, I suppose this issue can be closed.

andreasabel:
See #241: BNFC is now on stackage.

PR #199, opened 2016-11-24
#198 use stack to build bnfc

pascalh:
This pull request is associated to issue #198 and is just a preview, thus **not yet ready for merge**. However, building bnfc with stack and running the tests should be working (not all tests may succeed due to missing tool setup). Consider the pr as a basis for discussion.

Before merging this pull request in master the following things may need to be updated: 
- documentation
- travis CI


testing/scripts/runtests
```diff
@@ -1,5 +1,13 @@
...
+export PATH=$PATH:~/.local/bin
 export CLASSPATH=.:$(pwd)/data/javatools.jar
-mkdir -p dist/test
-cabal build
-cabal exec -- cabal run -- --xml=dist/test/bnfc-system-tests.xml $*
+cd ../source
+stack setup
+stack init
+stack test
+stack install
```
pascalh: I totally agree that tests should not have such a side effect. Let me give you a bit more context here: When developing on bnfc I felt like installing everything needed to run the system tests locally in order to make sure that I don't break anything. On the other hand I did not want to install a lot of new tools only for running system tests (especially haskell-platform). This is why I came up with a docker-based solution to run the system tests locally in a lightweight virtual machine (ubuntu) without installing all the dependencies on my main system. You can we my first steps using docker in this [https://github.com/pascalh/bnfc/tree/docker](branch). The cool thing about the docker based approach could be that the CI also just runs the docker image and thus there should be no difference between the local and the CI system tests. If you are interested in using that approach for system tests, I can also add a pull request for it.

source/BNFC.cabal
```diff
@@ -48,7 +48,7 @@ Executable bnfc
     array,
     process,
     containers,
-    pretty >=1.1 && <1.2,
```
gdetrez: I think that's the case only if stack is the only way to build the project. I'm not sure I want to go that far for different reasons: stack is still relatively new ; I'm not sure if it works well for package maintainer in gnu/linux distributions (they may have different versions of the dependencies) ; I think even the stack documentation recommends that, if you plan to upload your package to hackage, you should make sure it also builds with just cabal.

gdetrez:
> On the other hand we could setup the stack based build and talk about the system tests after that.

That's a good idea. I can already see advantages in being able to build bnfc with stack even without the system tests. We can see about running system tests with stack later, maybe once there is a new build server.

pascalh:
With this setup the following things work (and everything related to our old build using cabal is completely untouched):
```
stack setup 
stack build
stack test
cd testing/
stack exec bnfc-system-tests
```
I wanted to share this even if one has to enter the directory `testing` to run the system tests.

testing/scripts/runtests
pascalh: Just to clarify: in the branch I linked up there the system tests can be run in docker and locally. So the docker only adds a new possibility to run all the tests in a well configured environment. (But this can be discussed later on in another issue)

bafain:
Your commit c35ba97 ([#198]added bnfc-wide stack file) got rebased and merged into master as commit 3b6a007, with an updated LTS version. Please consider closing this pull request.

andreasabel:
Thanks, everyone!  I am closing this, as `stack-*.yaml` files are now present.

PR #201, opened 2017-01-11
Don't include <iostream> in C++

Teemperor:
The imported objects aren't used and there seems to be no
dependency on the includes of <iostream> in the code.

Also, including < iostream > in libraries is bad practice and some
projects completely forbid doing it (such as LLVM) because it
does some nasty static initializations.


andreasabel:
I am sorry you did not get a reply for your PR for almost a year.  There was simply no maintainer to look into this.

Would you be willing to rebase your PR onto the latest master?  I suppose the travis checks would then succeed.  
Another thing is to run the test suite (currently not automatic).  This is done as follows:
```
cd source
cabal install 
cd ../testing
cabal install
bnfc-system-tests
```
(assuming `.cabal/bin` is in your PATH).

andreasabel:
Thanks @Teemperor, I merged it now.

PR #218, opened 2017-12-18
Java Backend: Re-support cup version 0.11a

guardbotmk3:
commit 856d2db1ae79070bd15bd2c7e3fff18224265f79 accidentally pushed the required
version of cup to a fairly recent cup-0.11b cup-0.11a does not have offset
support in ComplexSymbol, so remove that for as long as we plan to support
cup-0.11a

### DIFF
--- a/source/src/BNFC/Backend/Java/CFtoJLex15.hs
+++ b/source/src/BNFC/Backend/Java/CFtoJLex15.hs
@@ -72,8 +72,7 @@ prelude (ln, jflex) packageBase = vcat
     , (if ln 
       then vcat
         [ "%line"
-        , (if jflex then "%column" else "")
-        , "%char" ]
+        , (if jflex then "%column" else "") ]
       else "")
     , "%public"
     , "%{"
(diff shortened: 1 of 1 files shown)


guardbotmk3:
Sorry about breaking the tests. I have a newer cup on my machine.

guardbotmk3:
The travis failure appears to be from a misconfiguration.

andreasabel:
Thanks!  As long as ubuntu does not provide 0.11b, I think it is better to stay on 0.11a.  I emailed the CUP maintainer on whether a release of 0.11b would be possible.
Concerning travis jobs, they seem often to fail for reasons out of our control.  Here, it seems like the company temporarily discontinued ghc 7.6.3 and 8.2.1.  Restarting the failed jobs often helps.
I restarted the failed jobs, and they seem to progress...

guardbotmk3:
Do you need anything else to merge this?

andreasabel:
Ah, no.  [I do not get notification for the passed travis test.]

PR #223, opened 2018-01-05
docs - add missing comment syntax information

danieljabailey:
Was looking for this information, took me a while to find it. A friend pointed out these docs with the crucial parts missing.

### DIFF
--- a/docs/lbnf.rst
+++ b/docs/lbnf.rst
@@ -963,8 +963,8 @@ The symbols used in BNF are the following::
 Comments
 --------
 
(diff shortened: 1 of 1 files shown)


andreasabel:
Thanks, I also noticed this.

I wondered whether this part of the .rst could be created from the lbnf grammar automatically.  A pandoc backend for bnfc would be something (instead of the latex backend).  Or a translation from the generated .tex to .rst via pandoc.

For now, let's take your fix.  Thanks!

Issue #221, opened 2018-01-02
Lists of lists


andreasabel:
Given this `.cf` file:
```
Start. S ::= [[E]] ;

Atom. E ::= Integer;
separator E "*"    ;
separator [E] "+"  ;
```
and this test input:
```
5 * 8 + 1 * 2 + 1
```
- The __haskell__ backend seems to produce a correct parser, but the printer drops the separators of the outer list:
```
happy -gca LL/Par.y
reduce/reduce conflicts: 3

[Abstract Syntax]
Start [[Atom 5,Atom 8],[Atom 1,Atom 2],[Atom 1]]

[Linearized tree]
5 * 8 1 * 2 1
```

- The __haskell-gadt__ backend cannot produce the abstract syntax:
```
happy -gca LL/Par.y
reduce/reduce conflicts: 3

[3 of 8] Compiling LL.Abs           ( LL/Abs.hs, LL/Abs.o )

LL/Abs.hs:21:55:
    Couldn't match type ‘Tree a0’ with ‘[E]’
    Expected type: Tree a0 -> m [E]
      Actual type: Tree a0 -> m (Tree a0)
    In the second argument of ‘(.)’, namely ‘f’
    In the second argument of ‘(.)’, namely ‘a (r (:)) . f’
```

- The __java__ backend cannot produce the abstract syntax:
```
LL/Absyn/ListListE.java:3: error: illegal start of type
public class ListListE extends java.util.LinkedList<[E]> 
```                                                    

- The __c__, __cpp__, and __cpp-nostl__ backends seem to work almost correctly; they seem to flatten lists when debug printing ASTs.
```
bison -t -pLL LL.y -o Parser.c
LL.y: warning: 2 reduce/reduce conflicts [-Wconflicts-rr]

[Abstract Syntax]
(Start [(Atom 5), (Atom 8), (Atom 1), (Atom 2), (Atom 1)])

[Linearized Tree]
5 * 8 + 1 * 2 + 1
```

- The __ocaml__ backend can't produce a printer:
```
ocamlyacc ParLL.mly
1 rule never reduced
3 reduce/reduce conflicts.

ocamlc  -o TestLL BNFC_Util.ml AbsLL.ml SkelLL.ml ShowLL.ml ...
File "ShowLL.ml", line 43, characters 66-74:
Error: This expression has type ('a -> showable) -> 'a list -> showable
       but an expression was expected of type ('a -> showable) -> showable
```

- The __latex__ backend works correctly.

[...]

PR #224, opened 2018-01-13
Absyn line tracking

guardbotmk3:
Adds support for line, column, offset tracking in the Abstract Syntax classes.
Because the location support in cup 0.11a is spotty, 0.11a is almost 12 years old, and location support didn't work until recently, I decided that for location support, we would require cup 0.11b-2014-06-11 or later. I verified that without the -l option, cup 0.11a can still process the resulting files.
2552ae9 contains some edits besides the reversion that allow it to work with 0.11a when line number support is off.

We'll have to work something out with travis to test the -l option only with a newish cup.


guardbotmk3:
Ping?

source/src/BNFC/Backend/Java.hs
```diff
@@ -352,7 +353,7 @@ cf2AntlrLex' l = CF2Lex
 
 -- | CF -> PARSER GENERATION TOOL BRIDGE
 -- | function translating the CF to an appropriate parser generation tool.
-type CF2ParserFunction = String -> String -> CF -> SymEnv -> String
+type CF2ParserFunction = String -> String -> CF -> Bool -> SymEnv -> String
```
andreasabel: Please don't use plain `Bool`. Define a 2-value data type with telling names instead.  Or at least define a type synonym.

source/src/BNFC/Backend/Java/CFtoAntlr4Parser.hs
```diff
@@ -53,8 +53,9 @@ type MetaVar     = (String, Cat)
 
 -- | Creates the ANTLR parser grammar for this CF.
 --The environment comes from CFtoAntlr4Lexer
-cf2AntlrParse :: String -> String -> CF -> SymEnv -> String
-cf2AntlrParse packageBase packageAbsyn cf env = unlines
+--The bool is line numbering, which is currently not supported.
```
andreasabel: Put argument documentations down as haddock comments, i.e, a `-- ^` comment following the argument type.  See https://github.com/andreasabel/haskell-style-guide/blob/master/haskell-style.md

andreasabel:
Thanks for the contribution!
One of my missions is to improve the code quality of bnfc, in terms of documentation, maintainability, automatic testing, etc.  Here is a style guide for Haskell code (WIP): https://github.com/andreasabel/haskell-style-guide/blob/master/haskell-style.md

guardbotmk3:
OK, Bool has been excised from the places where it was used previously, and the C++ Backend has been updated as well. I can change the name of the line numbering flag if you'd like, but at least it's not a boolean now.

andreasabel:
Thanks for the change.

Er, sorry to harp on this, but `LineNumber` sounds like you are passing a line number.  Could it be something like `HaveLineNumbers`, `HaveLocations`, `HaveLoc`, `HavePositions`, `HavePos` which indicates a flag?

guardbotmk3:
OK, I named it RecordPositions. I even renamed all the 'ln' variables rp. Thanks grep and sed.

guardbotmk3:
The Travis failure is the same one that I've seen before about GHC versions. The 2 non-broken versions passed.

guardbotmk3:
I re-ran the tests and encountered some errors, that I've cleaned up with this push. Everything should be in good order.

andreasabel:
Thanks, I merged this.
I resolved some conflicts with my own patches done for #221, which I pushed now.

PR #225, opened 2018-01-16
Use docker (to run system tests)

pascalh:
This pull request is about allowing to use bnfc inside of docker and especially running the system tests with a predefined container setup. Another benefit is that the current state of bnfc sources will be under test and not the installed version. 

For the moment I fixed the versions of the used packages (ghc, jdk, ...), but we can of course use variables to control the package version to install.


andreasabel:
Great!  How would I use this? Would you mind adding a readme with the instructions?

pascalh:
Sure, docker is properly set up on your machine if  `docker info` prints the docker system info.

Change to the directory containing the Dockerfile.

`docker build -t bnfc-system-tests . ` (This will setup a ubuntu container, download and install all the dependencies and bnfc.)

`docker run bnfc-system-tests` (This command will build and run the system tests.)

If you like, I can also add this to the `jenkins.yaml` to run the system tests on every build. This would then of course require to ignore the failing `cnf` tests, which I think is fine as we already have an issue for fixing them.

I can also add the commands to the readme.

andreasabel:
> I can also add the commands to the readme.

Yes, you can create a new readme under `testing/` maybe.

> If you like, I can also add this to the travis.yaml to run the system tests on every build. This would then of course require to ignore the failing cnf tests, which I think is fine as we already have an issue for fixing them.

That would be great!
On my machine, the `antlr` tests are also failing (in addition to the `cnf` tests).

andreasabel:
I merged your PR since you did work on the .cabal file.  
For now, I am trying to at least build the `testing` environment on travis, using the "build matrix" feature.
However, the ultimate goal is of course to also run the testsuite.

PR #226, opened 2018-01-18
Take advantage of JFlex's EOF handling for better error messages.

guardbotmk3:
This is relative to my other PR, So only look at e6908cd


andreasabel:
This PR does not build since it still uses `LineNumber`.

guardbotmk3:
Sorry, that's now fixed.

andreasabel:
Would you have a test case that demonstrates the better error message?

PR #229, opened 2018-06-23
Update lbnf.rst

dmjio:
Adjust `Exp` definition.

Without this change, the ADTs produced will be:

```haskell
data Exp = EPlus Exp Num | ENum Num
data Num = NOne
```

instead of
```haskell
data Exp = EPlus Exp Exp | ENum Num
data Num = NOne
```

### DIFF
--- a/docs/lbnf.rst
+++ b/docs/lbnf.rst
@@ -26,7 +26,7 @@ addition expressions with “1”:
 
 ::
 
-      EPlus. Exp ::= Exp "+" Num ;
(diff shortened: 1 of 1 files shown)


andreasabel:
Thanks for spotting this and your PR!  

I decided to fix it the other way round, since your proposed grammar is ambiguous.  There was another problem in the original example: `Num` will possibly conflict with Haskell's `Num`, thus I changed it to `Number`.

Issue #75, opened 2014-01-13
Options to choose makefile filename


robstewart57:
Currently the only option relating to generating a makefile is `-m`, which generates a makefile with filename `Makefile`.

Could an option be added to allow the user to also optionally specify the name of the makefile, e.g.

```
bnfc -m --make-filename=MyMakefile --haskell Foo.cf
```

PR #247, opened 2018-11-12
Distclean Makefile even when name is changed

bafain:
- `Common.Makefile.mkMakefile` now passes the specified makefile name to
  the code that generates the makefile content (f.i.,
  `Haskell.makefile`), which can use it to generate the appropriate `rm`
  command line for the `distclean` target

- The new test case distcleanTest checks that the test directory is
  empty except for the input file after a generate, build and distclean
  cycle

- See also
  - #75 (Options to choose makefile filename)
  - 883701c (Add an optional argument to change Makefile name)


source/src/BNFC/MultiView.hs
```diff
@@ -39,17 +39,14 @@ preprocessMCF f = do
 
 extract :: String -> LGrammar -> [(FilePath, Grammar)]
 extract name (LGr ldefs) =
-  [(file lang,Grammar [unldef ldef | ldef <- ldefs, isFor lang ldef]) |
+  [(file lang,Grammar $ concatMap (getDef lang) ldefs) |
```
andreasabel: Alternatively, `mapMaybe`http://hackage.haskell.org/package/base-4.12.0.0/docs/Data-Maybe.html#v:mapMaybe can be used for that purpose; then `getDef` would return either `Just d` or `Nothing`.

andreasabel:
Ready to merge?  (travis passes)

andreasabel:
There are already 3 different issues solved with this PR, I better hurry up and merge it before more gets in.

PR #288, opened 2020-01-26
[ CPP ] Throw runtime exception on parsing error

miketsukerman:
When BNFC generated code embedded into another project it's more convenient to get parsing errors as C++ exceptions rather than just via standard output.

I've added `parse_error` class to the `Absyn.H`
```
class parse_error : public std::runtime_error {
  public:
  parse_error(int line, std::string str) : std::runtime_error(str) , m_line(line) {}
  int getLine() { return m_line; }
  private:
  int m_line;
  };
```

### DIFF
--- a/source/src/BNFC/Backend/CPP/STL.hs
+++ b/source/src/BNFC/Backend/CPP/STL.hs
@@ -50,6 +50,7 @@ makeCppStl opts cf = do
     mkfile (name ++ ".y") bison
     let header = mkHeaderFile (inPackage opts) cf (allParserCats cf) (allEntryPoints cf) (Map.elems env)
     mkfile "Parser.H" header
+    mkfile "ParserError.H" (printParseErrHeader (inPackage opts) cf)
     let (skelH, skelC) = cf2CVisitSkel True (inPackage opts) cf
     mkfile "Skeleton.H" skelH
     mkfile "Skeleton.C" skelC
@@ -60,7 +61,35 @@ makeCppStl opts cf = do
     Makefile.mkMakefile opts $ makefile name
   where name = lang opts
 
-
+printParseErrHeader :: Maybe String -> CF -> String
+printParseErrHeader inPackage cf =
+  unlines
+  [
+     " #pragma once "
+     , " #include <string>"
+     , " #include <stdexcept>"
+     , ""
+     , nsStart inPackage
+     , " class parse_error : public std::runtime_error"
+     , " {"
+     , " public:"
+     , "     parse_error(int line, std::string str)"
+     , "         : std::runtime_error(str)"
+     , "         , m_line(line) {}"
+     , "     int getLine() {"
+     , "         return m_line;"
+     , "     } "
+     , " private:"
+     , "     int m_line;"
+     , " }; "
+     , nsEnd inPackage
+     ]
+     where
+      cat = head $ allEntryPoints cf
+      dat = identCat $ normCat cat
+      def = identCat cat
+      scope = nsScope inPackage
+   
 cpptest :: Maybe String -> CF -> String
 cpptest inPackage cf =
   unlines
@@ -71,11 +100,13 @@ cpptest inPackage cf =
     "/* pretty-print the result.                                                 */",
     "/*                                                                          */",
     "/****************************************************************************/",
-    "#include <stdio.h>",
-    "#include <string.h>",
+    "#include <cstdio>",
+    "#include <string>",
+    "#include <iostream>",
     "#include \"Parser.H\"",
     "#include \"Printer.H\"",
     "#include \"Absyn.H\"",
+    "#include \"ParserError.H\"",
     "",
     "void usage() {",
     "  printf(\"usage: Call with one of the following argument " ++
@@ -114,7 +145,12 @@ cpptest inPackage cf =
     "    }",
     "  } else input = stdin;",
     "  /* The default entry point is used. For other options see Parser.H */",
-    "  " ++ scope ++ dat ++ " *parse_tree = " ++ scope ++ "p" ++ def ++ "(input);",
+    "  " ++ scope ++ dat ++ " *parse_tree = nullptr;",
+    "  try { ",
+    "  parse_tree = " ++ scope ++ "p" ++ def ++ "(input);",
+    "  } catch( " ++ scope ++ "parse_error &e) {",
+    "     std::cout << \"Parse error on line \" << e.getLine(); ",
+    "  }",
     "  if (parse_tree)",
     "  {",
     "    printf(\"\\nParse Successful!\\n\");",
--- a/source/src/BNFC/Backend/CPP/STL/CFtoBisonSTL.hs
+++ b/source/src/BNFC/Backend/CPP/STL/CFtoBisonSTL.hs
@@ -106,6 +106,7 @@ header inPackage name cf = unlines
     , "#include <stdio.h>"
     , "#include <string.h>"
     , "#include <algorithm>"
+    , "#include \"ParserError.H\""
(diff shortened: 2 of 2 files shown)


miketsukerman:
Hi @andreasabel, I moved the definition of `parser_exception` to the separate header. Would it be okay?

miketsukerman:
Regarding the tests. To be honest, not that much proficient in Haskell. I would be very pleased if you could help me with tests or at least explain to me how to extend them ;)

andreasabel:
Thanks for the update, @miketsukerman !  At the moment I am a bit snowed under, but I will look at the PR when things are back to normal.

andreasabel:
Hi Mike,
with your patch the standard testsuite fails for `C++ (with namespace)`, e.g.:
```
[TEST] Parameterized tests:C++ (with namespace):Examples:LBNF
bnfc -mMyMakefile --cpp '-p foobar' LBNF.cf
make -f MyMakefile
error running: make -f MyMakefile
exit status: 2
stderr: LBNF.y:26:18: error: no member named 'parse_error' in namespace 'foobar'
  throw  foobar::parse_error( foobaryy_mylinenumber,str);
         ~~~~~~~~^
1 error generated.
```
One way to run the test is simply to type `make` in the bnfc root folder.

The generated `ParseError.H` typically looks like this:
```c++
 namespace bnfc
 {
 class parse_error : public std::runtime_error
 {
 public:
     parse_error(int line, std::string str)
         : std::runtime_error(str)
         , m_line(line) {}
...
```
But the call to `parse_error` in `Parser.C` looks like this:
```c++
throw foo::parse_error(fooyy_mylinenumber,str);
```

If `parse_error` would be placed into `Parser.H` I think this would work automatically, because `Parser.H` uses the namespace as given by the `-p` flag to BNFC.

However, you might have strong reasons to prefer a separate `.H` file.  Could you explain your rationale?

miketsukerman:
> Hi Mike,
> with your patch the standard testsuite fails for `C++ (with namespace)`, e.g.:
> 
> ```
> [TEST] Parameterized tests:C++ (with namespace):Examples:LBNF
> bnfc -mMyMakefile --cpp '-p foobar' LBNF.cf
> make -f MyMakefile
> error running: make -f MyMakefile
> exit status: 2
> stderr: LBNF.y:26:18: error: no member named 'parse_error' in namespace 'foobar'
>   throw  foobar::parse_error( foobaryy_mylinenumber,str);
>          ~~~~~~~~^
> 1 error generated.
> ```
> 
> One way to run the test is simply to type `make` in the bnfc root folder.
> 
> The generated `ParseError.H` typically looks like this:
> 
> ```c++
>  namespace bnfc
>  {
>  class parse_error : public std::runtime_error
>  {
>  public:
>      parse_error(int line, std::string str)
>          : std::runtime_error(str)
>          , m_line(line) {}
> ...
> ```
> 
> using its own namespace `bnfc`.
> 
> But the call to `parse_error` in `Parser.C` looks like this:
> 
> ```c++
> throw foo::parse_error(fooyy_mylinenumber,str);
> ```
> 
> expecting `parse_error` in the namespace as declared by the user.
> 
> If `parse_error` would be placed into `Parser.H` I think this would work automatically, because `Parser.H` uses the namespace as given by the `-p` flag to BNFC.
> 
> However, you might have strong reasons to prefer a separate `.H` file. (Then, of course, you would also have to use the namespace as given by the user.)
> Could you explain your rationale?

Very simple - just a mistake. Fixed

andreasabel:
Thanks.  Er, I meant the rationale of having a separate .H file.

miketsukerman:
> Thanks. Er, I meant the rationale of having a separate .H file.

It was the simplest way to include it in the `<bison>.y`

```
#include "ParserError.H"
#include "Absyn.H"

#define YYMAXDEPTH 10000000

typedef struct yy_buffer_state *YY_BUFFER_STATE;
int yyparse(void);
int yylex(void);
YY_BUFFER_STATE bnfcyy_scan_string(const char *str);
void bnfcyy_delete_buffer(YY_BUFFER_STATE buf);
int bnfcyy_mylinenumber;
void bnfcinitialize_lexer(FILE * inp);
int bnfcyywrap(void)
{
  return 1;
}
void bnfcyyerror(const char *str)
{
  extern char *bnfcyytext;
  throw bnfc::parse_error(bnfcyy_mylinenumber,str);
}
```

I can't include `Parser.H` here.

andreasabel:
Ok, I see, thanks for the explanation!

I ran the testsuite on your PR and now there is only a small problem left:
```
* Failures:
  * Parameterized tests:C++:distclean removes all generated files
  * Parameterized tests:C++ (with namespace):distclean removes all generated files
```
This means that the cleaning goals of the generated `Makefile` for the CPP-STL backend (the one you changed) need to be extended by the new file `ParserError.H`.

I can fix this myself.  Thanks for your contribution!

source/src/BNFC/Backend/CPP/STL.hs
```diff
@@ -114,7 +145,12 @@ cpptest inPackage cf =
     "    }",
     "  } else input = stdin;",
     "  /* The default entry point is used. For other options see Parser.H */",
-    "  " ++ scope ++ dat ++ " *parse_tree = " ++ scope ++ "p" ++ def ++ "(input);",
+    "  " ++ scope ++ dat ++ " *parse_tree = nullptr;",
```
andreasabel, line 148: Should have used `NULL` instead of `nullptr`.  Introduced regression #310.

source/src/BNFC/Backend/CPP/STL/CFtoBisonSTL.hs
```diff
@@ -124,8 +125,7 @@ header inPackage name cf = unlines
     , "void " ++ ns ++ "yyerror(const char *str)"
     , "{"
     , "  extern char *"++ns++"yytext;"
```
andreasabel, line 127: The declaration of `yytext` is now redundant and triggers a compiler warning on GNU C++ (not clang however):
```
test.y: In function 'void yyerror(const char*)':
test.y:25:16: warning: unused variable 'yytext' [-Wunused-variable]
   25 |   extern char *yytext;
      |                ^~~~~~
```

source/src/BNFC/Backend/CPP/STL.hs
```diff
@@ -114,7 +145,12 @@ cpptest inPackage cf =
...
+    "  try { ",
+    "  parse_tree = " ++ scope ++ "p" ++ def ++ "(input);",
+    "  } catch( " ++ scope ++ "parse_error &e) {",
+    "     std::cout << \"Parse error on line \" << e.getLine(); ",
```
andreasabel, line 152: Previously parse errors were reported on `stderr`, why now on `stdout`?

andreasabel (commented):
This is the review I should have done before merging, haha.

The patch introduced some regressions, unfortunately.

source/src/BNFC/Backend/CPP/STL/CFtoBisonSTL.hs
andreasabel: Fixed by 3b751a6dad594c6734c1474561a7db77ddd6af73.

source/src/BNFC/Backend/CPP/STL.hs
andreasabel: Fixed by 3b751a6dad594c6734c1474561a7db77ddd6af73.

source/src/BNFC/Backend/CPP/STL.hs
andreasabel: Fixed by 0415627efd93b209a664cd29885ca5b65a82afc5.

PR #310, opened 2020-10-29
Fix nullptr not declared error in CPP backend

felixonmars:
`nullptr` becomes a keyword since C++11. Let's use `NULL` instead.

```
g++ -g --ansi -W -Wall -Wno-unused-parameter -Wno-unused-function -Wno-unneeded-internal-declaration -c Test.C
Test.C:50:21: warning: identifier ‘nullptr’ is a keyword in C++11 [-Wc++11-compat]
   50 |   Exp *parse_tree = nullptr;
      |                     ^~~~~~~
Test.C: In function ‘int main(int, char**)’:
Test.C:50:21: error: ‘nullptr’ was not declared in this scope
At global scope:
cc1plus: note: unrecognized command-line option ‘-Wno-unneeded-internal-declaration’ may have been intended to silence earlier diagnostics
make: *** [Makefile:48: Test.o] Error 1
```

### DIFF
--- a/source/src/BNFC/Backend/CPP/STL.hs
+++ b/source/src/BNFC/Backend/CPP/STL.hs
@@ -140,7 +140,7 @@ cpptest inPackage cf =
     "    }",
     "  } else input = stdin;",
     "  /* The default entry point is used. For other options see Parser.H */",
-    "  " ++ scope ++ dat ++ " *parse_tree = nullptr;",
+    "  " ++ scope ++ dat ++ " *parse_tree = NULL;",
     "  try { ",
     "  parse_tree = " ++ scope ++ "p" ++ def ++ "(input);",
     "  } catch( " ++ scope ++ "parse_error &e) {",


andreasabel:
Thanks for reporting this bug.
This use of `nullptr` was introduced in recent commit ecccc61874812ab231e26559fb87777e5091d0a9 in PR #288.
Funnily, my C++ compiler does not complain here, must be Apple's fault.
```
$ g++ --version
Apple clang version 11.0.0 (clang-1100.0.33.17)
```
This is why I didn't notice this problem.

I wonder whether the better solution isn't to simply use `NULL` instead of `nullptr` and stay compatible with C++98.  Note that just a few lines above the problematic `*parse_tree = nullptr` there is the statement `char *filename = NULL`.  Thus, it is more consistent to continue to use `NULL`.

felixonmars:
Updated.

FTR, I am using the GNU toolchain:

```
$ g++ --version
g++ (GCC) 10.2.0
```

andreasabel:
Thanks for the update! I also installed the GNU compiler now and confirmed the problem and the fix.

Thanks for your contribution.  I fixed two more regressions of #288 along the way, see 3b751a6dad594c6734c1474561a7db77ddd6af73 .

PR #313, opened 2020-11-05
Update README.md

meemsbror:
I suspect the .cf file was omitted in the bnfc command for the examples for C-family backends.


andreasabel:
Thanks! 
Fixed in 74ac2de54a21c6f75cb72b2f9bc2454b6ee3cb9d .

Issue #314, opened 2020-11-06
C parser for a simple grammar does SEGV on Linux x86_64


varqox:
Encountered on version v2.8.4 (did not check others).
# How to reproduce
Create file `Instant.cf`:
```
Prog. Program ::= [Stmt] ;
SAss. Stmt ::= Ident "=" Exp;
SExp. Stmt ::= Exp ;
separator Stmt ";" ;

ExpAdd.            Exp1   ::= Exp2 "+"  Exp1 ;
ExpSub.            Exp2   ::= Exp2 "-"  Exp3 ;
ExpMul.            Exp3   ::= Exp3 "*"  Exp4 ;
ExpDiv.            Exp3   ::= Exp3 "/"  Exp4 ;
ExpLit.            Exp4   ::= Integer ;
ExpVar.            Exp4   ::= Ident ;
coercions Exp 4;
```
Compile:
```
bnfc --c --makefile Instant.cf && make
```
Run:
```
./TestInstant <<< "a=0"
```
Output:
```

Parse Successful!

[Abstract Syntax]
[1]    57036 segmentation fault  ./TestInstant <<< "a=0"
```
Expected output:
```

Parse Successful!

[Abstract Syntax]
(Prog [(SAss "a" (ExpLit 0))])

[Linearized Tree]
a = 0 

```

# What is going on?
I did some investigation and I think I know why this happens. First of all, it is printing that fails, because it tries to print `Ident` "a", but its value is invalid (for me it is 0x30 under GDB). This value comes from `strdup()`.
Now, during `make` the compiler gave us some warnings, which look like this:
```
Instant.l: In function ‘Instantlex’:
Instant.l:61:18: warning: implicit declaration of function ‘strdup’; did you mean ‘strcmp’? [-Wimplicit-function-declaration]
   61 | <YYINITIAL>{LETTER}{IDENT}*        yylval._string = strdup(yytext); return _IDENT_;
      |                  ^~~~~~
      |                  strcmp
Instant.l:61:16: warning: assignment to ‘char *’ from ‘int’ makes pointer from integer without a cast [-Wint-conversion]
   61 | <YYINITIAL>{LETTER}{IDENT}*        yylval._string = strdup(yytext); return _IDENT_;
      |                ^
Lexer.c: In function ‘Instant_init_buffer’:
Lexer.c:1656:48: warning: implicit declaration of function ‘fileno’ [-Wimplicit-function-declaration]
 1656 | 
      |                                                ^     
```
[...]

PR #315, opened 2020-11-06
Fix invalid Ident values in C backend

varqox:
Proposed fix for #314.


andreasabel:
I tried to fix this by including the definition of `_POSIX_C_SOURCE` in the `.l`file, but it is inserted into `Lexer.c` too late to have effect (after the inclusion of `string.h`).
So I am merging your fix, also in the light that `strdup` seems to be moving into the standard: https://en.cppreference.com/w/c/experimental/dynamic/strdup

PR #318, opened 2020-11-10
Add newline to Parse error message in C++ backend

meemsbror:
There was a missing newline on the error message where the parsing failed


andreasabel:
Thanks!

PR #321, opened 2020-11-12
[ fixed #319 ] proper escaping in ANTLR-Lexer definition

### LINKED ISSUE #319: Antlr backend - quotation marks in bracket expressions are escaped when they shouldn't
It seems the only characters that should be escaped in bracket expressions in regexes are `]`, `\`, and `-`. I'm not sure if this means that there needs to be different escaping in different contexts.
https://github.com/antlr/antlr4/blob/master/doc/lexer-rules.md#lexer-rule-elements

Example token rule that generates broken code (not by any means good or correct, I just noticed that the resulting lexer file doesn't work) :
`token NoteToken ["abcdefgr"]({"es"} | {"is"})*["\',"]*(digit)*["."]* ;`
results in the following line in the Lexer.g4 file
`NoteToken : [abcdefgr]('e''s'|'i''s')*[\',]*DIGIT*'.'*;`
which generates the following when building:
`warning(156): lily/lilyLexer.g4:83:38: invalid escape sequence \'`

The build also complains about the following line:
`STRINGTEXT : ~[\"\\] -> more;`

https://github.com/BNFC/bnfc/blob/3ca72116c4a8f541dff2778450efabd72219cf8e/source/src/BNFC/Backend/Java/CFtoAntlr4Lexer.hs#L157

The build works as expected when removing the extra backslashes as follows:
`NoteToken : [abcdefgr]('e''s'|'i''s')*[',]*DIGIT*'.'*;`
...
`STRINGTEXT : ~["\\] -> more;`

----

Sidenote:
I first thought this could be related to this line, referencing RegToJLex.hs instead of RegToAntlrLexer.hs, but it seems the reference is correct, even if it's confusing naming.
https://github.com/BNFC/bnfc/blob/3ca72116c4a8f541dff2778450efabd72219cf8e/source/src/BNFC/Backend/Java/CFtoAntlr4Lexer.hs#L150 

Export from RegToAntlrLexer:
https://github.com/BNFC/bnfc/blob/3ca72116c4a8f541dff2778450efabd72219cf8e/source/src/BNFC/Backend/Java/RegToAntlrLexer.hs#L1

andreasabel:
Inside [...], there is no need to escape single quotes.

Also fixed a problem with missing spaces in sequences, e.g.

  token Foo letter letter

produced LETTERLETTER.

### DIFF
--- a/source/src/BNFC/Backend/Java/CFtoAntlr4Lexer.hs
+++ b/source/src/BNFC/Backend/Java/CFtoAntlr4Lexer.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE NoImplicitPrelude #-}
+{-# LANGUAGE OverloadedStrings #-}
 
 {-
     BNF Converter: Java Antlr4 Lexer generator
@@ -91,7 +92,7 @@ cMacros = vcat
   where frg a = "fragment" <+> a <+> ";"
 
 escapeChars :: String -> String
-escapeChars = concatMap escapeChar
+escapeChars = concatMap escapeCharInSingleQuotes
 
 -- |
 -- >>> lexSymbols [("foo","bar")]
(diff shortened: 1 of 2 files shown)


source/src/BNFC/Backend/Java/CFtoAntlr4Lexer.hs
```diff
@@ -154,7 +155,7 @@ restOfLexerGrammar cf = vcat
         , "mode STRINGMODE;"
         , "STRINGESC : '\\\\' -> more , pushMode(STRESCAPE);"
         , "STRINGEND : '\"' ->  type(STRING), mode(DEFAULT_MODE);"
-        , "STRINGTEXT : ~[\\\"\\\\] -> more;"
+        , "STRINGTEXT : ~[\"\\\\] -> more;"
         ]
     charmodes     = [ "mode CHARMODE;"
         , "CHARANY     :  ~[\\'\\\\] -> more, mode(CHAREND);"
```
andreasabel, line 161: Here was another wrongly escaped quote, see #329.

PR #327, opened 2020-12-07
Fix #176: Store node position in the AST (Haskell backend)

### LINKED ISSUE #176: Question: Using functor to store position information
According to `bnfc --help`:
`--functor       Make the AST a functor and use it to store the position of the nodes`

How would you go about storing the position of the nodes?
The type that

``` haskell
parseProgram = pProgram . myLexer
```

yields is:

``` haskell
parseProgram :: String -> Err (Program ())
```

I.e: no position information is added... So does that mean that you have to use something else than `happy`, or can you get it to add that position information for you?

Commelina:
Most of the code is taken and reorganized from the unmerged branch https://github.com/BNFC/bnfc/compare/176-source-position. I tested it on all three token types with/without functor. But I am not sure if it overrides some new features when generating Happy file.


andreasabel:
Great, @Commelina to work on this dearly needed feature!

I build your PR and ran the `bnfc-system-tests` (directory `testing`).

> But I am not sure if it overrides some new features when generating Happy file.

Unfortunately, it seems to.  Please have a look at the following tests:

Regression tests:
- [ ] 266_define
- [ ] 235_SymbolsOverlapTokens
- [ ] 70_WhiteSpaceSeparator
- [ ] 202_comments
- [ ] 204_InternalToken

Examples:
- [ ] Alfa
- [ ] cubicaltt

Hint: in `testing/src/ParameterizedTests.hs`, move the test parameters for `--haskell --functor` first in the list.
https://github.com/BNFC/bnfc/blob/e7e68ba84c20b662f61762b9d054bbd9a175c8de/testing/src/ParameterizedTests.hs#L332-L333
If you then `make` in the `testing` directory, the relevant tests run first (the whole testsuite takes 1 hour).

Commelina:
Hi, I have rewritten the module based on the latest code from the `master` branch instead of modifying branch `176-source-position`. I think that it is much clearer than the unmerged branch, and minimizes the occurrence of overriding new features. Only `CFtoHappy.hs` is modified instead of adding extra modules in this version.

Then I ran the `bnfc-system-tests` (in fact the part related to Haskell) and compared the output with the one from the latest unmodified code. The result is exactly the same, which means if the unmodified version passes a test, the modified one also passes and vise versa (in fact there does exist the situation that the unmodified code fails to pass some of the tests).

Also, I tested it on my own projects and it works well.

source/src/BNFC/Backend/Haskell/CFtoHappy.hs
```diff
@@ -231,36 +257,65 @@ footer = unlines $
...
+            ++++ own ++ "  : L_" ++ own ++ " { " ++ mkBodyPart t ++ " }"
   where
+    mkTypePart tokenCat = if functor then "(Maybe (Int, Int), " ++ tokenCat ++ ")" else tokenCat
+    mkBodyPart tokenCat
+      | null mkPosPart = mkValPart tokenCat
+      | otherwise      = "(" ++ mkPosPart ++ ", " ++ mkValPart tokenCat ++ ")"
+    mkPosPart = if functor then "Just (tokenLineCol $1)" else ""
+    mkValPart tokenCat =
+      case tokenCat of
+        "String"  -> if functor then stringUnpack "(tokenText $1)"
```
andreasabel, line 301: Testcase `235_SymbolsOverlapTokens` is broken: `String` literals now arrive quoted in the abstract syntax.  This is due to the use of `tokenText` here, I suppose.  Note that  the generated `tokenText` for `String` tokens uses the `show` function, which adds a level of quotation.

source/src/BNFC/Backend/Haskell/CFtoHappy.hs
```diff
@@ -133,17 +141,31 @@ rulesForHappy absM functor cf = for (ruleGroups cf) $ \ (cat, rules) ->
...
+           | functor                        = getFunctorAction
            | otherwise                      = unwords (qualify fun : metavars)
+    -- Commelina, 2020-12-10:
+    -- replace the previous unimplemented "()"
+    getFunctorPos rhs = case rhs of
+      []          -> "Nothing"
+      (Left _:_)  -> "fst $1"
+      (Right _:_) -> "Just (tokenLineCol $1)"
+    getFunctorValue
+      | isCoercion fun = unwords metavars
```
andreasabel, line 162: This duplicates some code of `action`.  Please refactor so that no code is duplicated.

andreasabel (changes requested):
Thanks, @Commelina, this version is very good!

I noticed some problems with `String` literals (testcase `235_...`), see below.

Further, please avoid to introduce code duplication.  (The codebase of BNFC unfortunately suffers a lot from code duplication.  My goal is to get rid of duplication as much as possible.)  Maybe you can even find opportunities to reduce duplication in `CFtoHappy` in the places you modify.

source/src/BNFC/Backend/Haskell/CFtoHappy.hs
```diff
@@ -133,17 +141,31 @@ rulesForHappy absM functor cf = for (ruleGroups cf) $ \ (cat, rules) ->
...
+           | isNilCons  fun && not functor  = unwords (qualify fun : metavars)
+           | functor                        = getFunctorAction
            | otherwise                      = unwords (qualify fun : metavars)
+    -- Commelina, 2020-12-10:
+    -- replace the previous unimplemented "()"
+    getFunctorPos rhs = case rhs of
+      []          -> "Nothing"
+      (Left _:_)  -> "fst $1"
+      (Right _:_) -> "Just (tokenLineCol $1)"
+    getFunctorValue
```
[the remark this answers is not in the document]
Commelina: Maybe it refers to `getFunctorValue`...? But indeed, the code needs refactoring.

After refactoring, the hunk contains three core functions `actionPos`, `actionValue` and `action`. And now `actionValue` can handle both situations with/without functor so there is less duplication. `actionPos` gets position information and `action` just composes them in different forms depends on `functor`:

```haskell
    action
      | functor   = let pos = actionPos
                        val = actionValue
                     in "(" ++ pos ++ ", " ++ val ++ ")"
      | otherwise = actionValue
    actionPos = case rhs of
      []          -> "Nothing"
      (Left _:_)  -> "fst $1"
      (Right _:_) -> "Just (tokenLineCol $1)"
    actionValue
      | isCoercion fun = unwords metavars
      | isNilCons  fun = unwords (qualify fun : metavars)
      | functor        = unwords (qualify fun : ("(" ++ actionPos ++ ")") : metavars)
      | otherwise      = unwords (qualify fun : metavars)
```

source/src/BNFC/Backend/Haskell/CFtoHappy.hs
Commelina: That is my mistake. Indeed, the `show` function in `tokenTest` adds an extra level of quotation to `String` literals. 

The simplest and clearest way I can find out is to replace the `tokenText` function with a partial one `(\(PT _ (TL s)) -> s)`:

```haskell
"String"  -> if functor then stringUnpack "(\\(PT _ (TL s)) -> s) $1" 
                        else stringUnpack "$1"
```

Although the function is partial, there will never be a `Tok` other than `TL s` when `"String"` is matched. So it is safe, I think. 

(And I finally find out why my testing gave me unexpected results because I forgot to replace `bnfc` with the latest build :( )

source/src/BNFC/Backend/Haskell/CFtoHappy.hs
Commelina: I tried resolving this problem in the first conversation (https://github.com/BNFC/bnfc/pull/327#discussion_r541956536).

andreasabel:
@Commelina thanks for the fixes!

The testsuite passes now.

We can now look at the user experience (UX).  I tried `--functor` on a tiny grammar for Hutton's razor:
```
EPlus. Exp  ::= Exp "+" Exp1;
EInt.  Exp1 ::= Integer;
```
Then I looked at the products to discover how to use the new feature.  
- `Test.hs` is unchanged w.r.t. the version without `--functor`.  This is good.  Yet it did not give me any hints how to use positions.
- `Abs.hs` places an additional parametric value at each AST constructor
  ```haskell
  data Exp a = EInt a Integer | EPlus a (Exp a) (Exp a)
    deriving (C.Eq, C.Ord, C.Show, C.Read)
  
  instance C.Functor Exp where
  ```
  I think the `Functor` instance can and should now be automatically generated by GHC using the `DeriveFunctor` extension. Also, `Foldable` and `Traversable` can be derived.
  The file `Abs.hs` should give me some clues about the positions that the parser delivers.  I'd favor something along
  ```haskell
  data Position = Position { startLine :: Int, startColumn :: Int }
  data Exp' a = EInt a Integer | EPlus a (Exp' a) (Exp' a) ...
  type Exp = Exp' Position
  ```
- `Par.y`:
  ```haskell
  %name pExp_internal Exp
  %name pExp1_internal Exp1
  ...
  %%

  Integer :: { (Maybe (Int, Int), Integer) }
  Integer  : L_integ  { (Just (tokenLineCol $1), (read ((tokenText $1))) :: Integer) }

  Exp :: { (Maybe (Int, Int),  (Razor.Abs.Exp (Maybe (Int, Int))) ) }
  Exp : Exp '+' Exp1 { (fst $1, Razor.Abs.EPlus (fst $1) (snd $1) (snd $3)) }

  Exp1 :: { (Maybe (Int, Int),  Razor.Abs.Exp (Maybe (Int, Int)) ) }
  Exp1 : Integer { (fst $1, Razor.Abs.EInt (fst $1) (snd $1)) }
  {
  ...
  myLexer = tokens
  pExp = (>>= return . snd) . pExp_internal
  pExp1 = (>>= return . snd) . pExp1_internal
  }
  ```

Issues with the generated parser:
- Concerning the last lines: The simpler code is
  ```haskell
  pExp = fmap snd . pExp_internal
  ```
- `pExp` etc. should have a type signature
- There should be a comment to point out the entrypoints at the end of the file (for better UX).
- (Minor issue, not sure: `pExp_internal`:  better name scheme would be `pExpWithPosition` etc.)
- The `Maybe (Int, Int)` shouldn't be duplicated everywhere.  See comment on `Abs.hs` how to factor out a type of `Position` and types for AST trees with position information.  
- Can we get `Position` instead of `Maybe Position`?  I see the `Nothing` comes from the empty list case, but maybe there we could return the position of the following token?

From a long term maintenance perspective, making the right choices concerning the user interface to the new `--functor` feature is important, as these will have to stay fixed for reasons of backwards compatibility.  This is why we should think carefully.  Maybe this feature should go into 2.9.1 rather than 2.9.0 to give it some time to mature.

I see the status of this PR as "can be merged", but further work is needed to smoothen the UX.
What are your plans?  Do you want to continue here to also engineer the UX aspects, or do you want this PR merged and be done with it?

Commelina:
Thanks for your detailed reply! 

The items you listed are worthy of improvement, and I'd like to continue here. However, I am afraid that I do not have enough time in the following few days (maybe it will be better after a few weeks). In my personal opinion, to avoid conflicts in the future, I'd like to merge it now and improve it later. I will improve them as soon as I have time.

andreasabel:
Ok, I will try to release 2.9.0 soon, then merge this PR, then look at UX design.

andreasabel:
@Commelina : Your PR has been merged as commit 6d57125c999271bff5d4cda6fd62c945abc828fa.
Thanks for your contribution!

andreasabel:
> ```haskell
> data Position = Position { startLine :: Int, startColumn :: Int }
> ```

I implemented something like this, but I have second thoughts because of _generativity_.  If you have two grammars, BNFC will create two `Position` types. Albeit they are isomorphic, they are not the _same_ to the compiler, and if you add some tools to manipulate positions, you need to coerce one of these types into the other.  A simple type synonym
```
type Position = Maybe (Int, Int)
```
does not have these problems.

PR #351, opened 2021-03-24
[ fix #349 ] Make parser reentrant by using reentrant lexer

### LINKED ISSUE #349: Make C parser reentrant
The `*.y` template file generated by BNFC using global varilables to store the parsed AST and internal status.

```c
/* Global variables holding parse results for entrypoints. */
Proc YY_RESULT_Proc_ = 0;
ListProc YY_RESULT_ListProc_ = 0;
```

This means the parser is not thread-safe.

Please consider enabling Thread Local Storage for them.
Here is a `thread_local` macro suggested [here](https://stackoverflow.com/questions/18298280/how-to-declare-a-variable-as-thread-local-portably).

```c
#ifndef thread_local
# if __STDC_VERSION__ >= 201112 && !defined __STDC_NO_THREADS__
#  define thread_local _Thread_local
# elif defined _WIN32 && ( \
       defined _MSC_VER || \
       defined __ICL || \
       defined __DMC__ || \
       defined __BORLANDC__ )
#  define thread_local __declspec(thread) 
/* note that ICC (linux) and Clang are covered by __GNUC__ */
# elif defined __GNUC__ || \
       defined __SUNPRO_C || \
       defined __xlC__
#  define thread_local __thread
# else
#  error "Cannot define thread_local"
# endif
#endif
```

andreasabel:
[ fix #349 ] Make parser reentrant by using reentrant lexer

Further:

- let bison generate Bison.h, to slim down Parser.h
    
    Parser.h defined things that are only for internal use in the lexer.
    These can be automatically defined by bison using the %defines pragma.

- C parser: return parse result in parameter to yyparse
    
    This is more robust than using global variables, prepares for making
    parser reentrant.

- C parser: code segment after %union, entrypoints in the end
    
    Reorganized the input for bison a bit, to prepare for removal of
    global YY_RESULT variables.
    
    Got rid of the forward declaration for yyerror.
    
    Entrypoints can be at the end of the file, since nothing depends on them.

- C lexer: use flex's predefined start state INITIAL
    
    Use INITIAL instead of defining our own YYINITIAL.


wangjia184:
Tested in linux/x64/gcc  and macos/arm64/clang
neither works

random errors when run my test cases  in parallel, should I add mutex to sync the test case?

andreasabel:
I realized there is still a global variable in the lexer for lexing strings, it is `literal_buffer`.  I have to make this local as well...

wangjia184:
@andreasabel still same error.
In generated files, I see change like this one.
image

Just make sure I updated my BNFC

andreasabel:
@wangjia184 :  Thanks, yes this was the last change I did.

I am a bit out of ideas now how to proceed, since I don't have some infrastructure for running the parser in parallel.
- Could you share a small test framework with me where I can debug the parallelism problem myself?
- Alternatively, could you find out how to change the output of BNFC, in particular the generated `.l` and `.y` files, so that the parser works in parallel?  I could then do the modification of BNFC accordingly.

andreasabel:
This project spilled over to the C++ backends.  Since the work on the C backend broke the C++ backends, I took the forward route to (re)integrate the C++ lexer/parser generators into the C backend.  This way, we also get reentrant parsers for C++.

PR #357, opened 2021-04-23
[ C ] bufferAppendString: Copy in the trailing 0 with the string.

shlevy:
Fixes Wstringop-truncation on gcc 10.2.0


source/src/BNFC/Backend/C.hs
```diff
@@ -348,12 +348,9 @@ bufferC bufferH = unlines
   , "    /* Make sure the buffer can hold all of s. */"
   , "    bufferAllocateChars(buffer, len);"
```
shlevy, line 349: I'm actually not clear if this is right, though it's not a *new* bug if so. Does `bufferAllocateChars` include a + 1 for the nul?

andreasabel: The buffer `size` is always `>= 1` to hold at least the terminating null.  `bufferAllocateChars` makes sure that `len` extra characters have space.  There is no need to include the null in the count `len` because there is already room for the terminating null.
```c 
static void bufferAllocateChars (Buffer buffer, const unsigned int n) {
  /* 1 extra char for terminating 0. */
  unsigned int requiredSize = buffer->current + 1 + n;
...
```
TL;DR: Yes, it includes the +1, as you see.

source/src/BNFC/Backend/C.hs
```diff
@@ -348,12 +348,9 @@ bufferC bufferH = unlines
...
   , ""
-  , "    /* Append s at the end of the buffer. */"
-  , "    strncpy(buffer->chars + buffer->current, s, len);"
```
andreasabel, line 352: LGTM, I didn't know you can also copy the terminating null, but so be it.

andreasabel (approved):

andreasabel:
For information: did you test this, e.g. one one test parser or with `bnfc-system-tests`?

shlevy:
@andreasabel Not yet, didn't get the build system up. Happy to put this on hold until we do (probably later today). I did test that this compiles when changing the generated C manually.

andreasabel:
Works for me.

Issue #363, opened 2021-04-28
Make Ulf Norell's `define` example work in all backends


andreasabel:
The original example for `define` by Ulf Norell is currently only accepted by the Haskell backends.
https://github.com/BNFC/bnfc/blob/3da1b0e732bd67b264d1783f6804c6b101147401/examples/DefinedRules.cf#L22-L23
- [x] Agda: wrong list constructors
- [x] C: list constructors rendered wrongly as `make_(:)` and `make_[]`.
- [x] CPP: missing sanitization of `if` and `for` (keyword clash, regression), wrong list constructors.
- [x] Ocaml: missing sanitization (regression), list cons `(:) (a,b)` instead of `(a :: b)`.
- [x] Java: missing sanitization ([regression](https://github.com/BNFC/bnfc/commit/6349a6137971f6a5665ee51a8e7e8bdb493ca5b1#r50133814)), complaint that `if` clashes with `If`.  Works in 2.9.1 with `--force`.

Missing sanitization is a regression wrt. 2.9.1, introduced by #287.
The list constructors probably never worked.

shlevy:
@andreasabel We just hit:

> C: list constructors rendered wrongly as make_(:) and make_[]

Any pointers to fix?

andreasabel:
I am now working to extend the expression type by typing information, so that the printers can pick the right list constructors.

andreasabel:
Remaining problem (C): non-linear definitions, like:
https://github.com/BNFC/bnfc/blob/228e2eff0f4f4269b8d375814fc544c506ea6e2c/examples/define/test.cf#L25
This creates syntax trees that not trees but DAGs.
```
  make_Assign (x, make_EOp (make_EVar(x), make_Plus(), make_EInt(1)))
```
E.g. here, `x` is shared.  This causes problems for the C deallocator
which assumes that it is dealing with a tree.

shlevy:
> the C deallocator

A bit off topic, but is there a deallocation function provided with the generated files? I haven't seen one, I've just been manually freeing recursively down the generated structs.

andreasabel:
@shlevy : The deallocation is a new, yet unreleased functionality added in a9d360d9cc0a33624e3d7f3e9c8cda01ab25cb9c (#348).

andreasabel:
> Remaining problem (C): non-linear definitions, like:

This has been fixed by introduction of `clone_*` and the cloning of duplicate components to prevent pointer aliasing and to preserve the tree-property.

PR #365, opened 2021-04-30
C: Properly implement defined rules using the empty list.

shlevy:
Part of #363.

### DIFF
--- a/source/src/BNFC/Backend/C/CFtoCAbs.hs
+++ b/source/src/BNFC/Backend/C/CFtoCAbs.hs
@@ -126,6 +126,8 @@ prDefH tokenCats f xs e = concat [ "#define make_", f, "(", intercalate "," xs, 
     -- Token categories are just @typedef@s in C, so no constructor needed.
     App g [e] | g `elem` tokenCats
                 -> prExp e
+    App "[]"  [] -> "NULL"
+    App "(:)" [hd, tl] -> error "nonempty lists in defined rules not supported in C" -- How do we get the type of list at this point?
     App g es    -> concat [ "make_", g, "(", intercalate "," (map prExp es), ")" ]
     LitInt    i -> show i
     LitDouble d -> show d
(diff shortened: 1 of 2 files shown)


andreasabel:
Thanks.
My suggestion would be to generalize `Exp` in `CF.hs` to
```haskell
type Exp = Exp' String
data Exp' f
  = App f [Exp]       -- ^ (Possibly defined) label applied to expressions.
  | Var String        -- ^ Function parameter.
  | LitInt Integer
  | LitDouble Double
  | LitChar Char
  | LitString String
  deriving (Eq, Functor)
```
Then replace, in the C backend, `[]` by `Null` and `(:)` by the respective `make_ListXXX`.  (Btw. there is `isNilFun` and `isConsFun` to check for these special labels.)
This is likely still not enough, we need to have the list type `XXX`.  This could be saved in the `TypeChecker.hs`, producing a `Exp' (String,Cat)` remembering the target type of the `App f es`.  Spinning this further, the `Cat` is only needed for list constructors, so maybe `Exp' Label` would be sufficient where
```haskell
data Label
  = OrdinaryLabel String
  | NilLabel  Cat    -- ^ []
  | ConsLabel Cat   -- ^ (:)
  | OneLabel  Cat    -- ^ (:[])
```

shlevy:
Makes sense, will try to carve out some time!

andreasabel:
I fixed this problem in a way similar as I wrote in the comment.
Mainly, we remember the typing information from the type checker, so we
- do not need to run the type checker again (was so in the OO backends) when translating the definitions
- can use the typing information to select the right list constructors.

shlevy:
Awesome! Sorry this sat for so long, I only have a little time here and there to work on the project using BNFC at the moment. Hopefully will be able to pick up the rust backend soon!

andreasabel:
Yeah, looking forward to the Rust backend!  

If you find any more bugs, do not hesitate to report them!  (Trying to understand how things work is the best way to discover bugs...)

PR #382, opened 2021-09-15
C backend: Parser.h: #include <stdio.h>

### LINKED ISSUE #381: C backend assumes that user of header file includes `stdio.h`
In the C backend, none of the generated headers include any header from standard library. This is wrong, as `Parser.h` uses `FILE*`, which is defined in `stdio.h`. Thus, `stdio.h` should be also included.

BNFC version that I use: 2.9.1.
OS: Fedora 34
BNFC distribution: `stack`

enedil:
Parser.h uses FILE*, however doesn't include stdio.h, even transitively.
This is an issue if using the C backend from Zig, where usage of C FILE
api is not needed.

Fixes #381.

### DIFF
--- a/source/src/BNFC/Backend/C.hs
+++ b/source/src/BNFC/Backend/C.hs
@@ -220,6 +220,7 @@ mkHeaderFile _ cf _env = unlines $ concat
   [ [ "#ifndef PARSER_HEADER_FILE"
     , "#define PARSER_HEADER_FILE"
     , ""
+    , "#include <stdio.h>"
     , "#include \"Absyn.h\""
     , ""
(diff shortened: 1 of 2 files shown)


andreasabel:
Thanks for adding to `STL.hs`, would it be possible you also updated `NoSTL.hs`?

enedil:
> Thanks for adding to `STL.hs`, would it be possible you also updated `NoSTL.hs`?

I see you already did it. Thanks.

andreasabel:
@enedil: Yes, as I anyway had to test on my local machine, I thought it is easier that I do it myself.

PR #385, opened 2021-09-22
Fix #384: Haskell/OCaml printer: nil case also for nonempty lists

### LINKED ISSUE #384: Printer for nonempty lists: "pattern matching is not exhaustive" warning in ocaml backend (suppressed in Haskell backend)
Background: I want to have a list of ``Data`` category elements with at least two elements. The easiest working solution I found is
```
DPair. Data ::= "Pair" Pair ;
DPairSeq1.   Pair ::= Data Data ;
DPairSeq2.   Pair ::= Data Pair ;
```

However, here are two other ways i thought of, that cause a "pattern matching is not exhaustive" warning in the ocaml / ocaml-menhir backend:
```
DPair. Data ::= "Pair" Pair ;
terminator nonempty Pair "" ;
-- DPairSeq1.   Pair ::= Data Data ;
DPairSeq2. Pair ::= Data [Pair] ;
```

The full error message:
```
...
ocamllex LexMichelson.mll
28 states, 718 transitions, table size 3040 bytes
ocamlc  -o TestTest BNFC_Util.ml AbsTest.ml SkelTest.ml ShowTest.ml PrintTest.ml ParTest.mli ParTest.ml LexTest.ml TestTest.ml
File "PrintTest.ml", lines 91-93, characters 33-63:
91 | .................................match (i, es) with
92 |     (_,[x]) -> (concatD [prtPair 0 x])
93 |   | (_,x::xs) -> (concatD [prtPair 0 x ; prtPairListBNFC 0 xs])
Warning 8: this pattern-matching is not exhaustive.
Here is an example of a case that is not matched:
(_, [])
```

The second gramar that causes the same error:
```
DPair. Data ::= "Pair" Data [Pair] ;
terminator nonempty Pair "" ;
DPairSeq2. Pair ::= Data ;
```

The warning does not occur if ``nonempty`` is removed.

andreasabel:
Fix #384: Haskell/OCaml printer: nil case also for nonempty lists

Not 100% sure whether I want to merge this, as it only fixes the `nonempty` list case, not the precedence category list case (e.g. `[Exp2]`).


andreasabel:
Update: also fixing the precedence list issue.  So, fine to merge.

Issue #293, opened 2020-04-29
Option to not use `new` in C++


sorenbug:
I would love it if there was an option on BNFC to parse into either a `std::shared_ptr` or to stack allocate classes because it's been a chore to track down everything that's been parsed and properly `delete` it.

andreasabel:
I am not proficient enough in C++ to attempt an implementation, but PRs are welcome!

sorenbug:
I don't know that much about the tools BNFC uses to generate C++ code, but I'll do my best.

andreasabel:
BNFC does not use a lot of tools; it simply prints concrete syntax, maybe with the help of the `Text.Pretty` pretty printing library.

bbyalcinkaya:
Makefiles for C++ backend have `--ansi` flag and C++98 does not support `shared_ptr`/`unique_ptr`. Is it safe to switch to C++11?

andreasabel:
> Makefiles for C++ backend have `--ansi` flag and C++98 does not support `shared_ptr`/`unique_ptr`. Is it safe to switch to C++11?

Well, if there is an option
> to parse into either a std::shared_ptr/std::unique_ptr

then this option would output a `Makefile` with the appropriate flags (like picking the necessary C++ version).

So, short answer: yes.

hangingman:
@andreasabel I created second pull-request at #410
I gave up using `std::unique_ptr`, then use `std::shared_ptr`. Because it's more applicable for bnfc usecase.
Please take a look it.

hangingman:
@andreasabel Hi, I updated PR again.
It took much time, sorry.
https://github.com/BNFC/bnfc/pull/410

hangingman:
@andreasabel 
Hi, long time no see.
Is there a chance that my pull request (#410) will be merged?
If regression is a problem, I can divide the pull request into small ones or make a pull request to be test functions only.

PR #389, opened 2021-10-09
[WIP] Issues/293 use new c++ feature and some fix

hangingman:
ref #293

PR description:
I modified some C++ code related with issue #293 . ~This is not so big change.~

### PR contents

- Makefile changes:
  - If bnfc received `--ansi` flag, Makefile would have `-std=c++14` compile option otherwise `--ansi`.
  - If bnfc not received `--ansi` flag, bnfc will generate `.ll` and `.yy` files

- Fix PARSER_HEADER_FILE
  - C++ source should have same name interface like "psProgram" in header file and source file. But, currently it looks Parser.H doesn't match function name.

- C++ AST/Test code generator switching
  - If bnfc not received `--ansi` flag, bnfc would genrate codes using smart-pointer (std::unique_ptr), otherwise bnfc generate codes same as before.

Sample code: 

use `using` instead of `typedef`

```cpp
/********************   TypeDef Section    ********************/

using Integer = int;
using Char = char;
using Double = double;
using String = std::string;
using Ident = std::string;
```

`clone()` returns instance wrapped `std::unique_ptr`
```cpp
/********************   Abstract Syntax Classes    ********************/

class Program : public Visitable
{
public:
  virtual std::unique_ptr<Program> clone() const = 0;
};
```

- Impl classes have each copy constructors
  - I referred the book "More Effective C++" to implement it.
```cpp
class Prog : public Program
{
private:
  std::unique_ptr<ListStatement> liststatement_;

public:
  // for "std::move"
  Prog(Prog&& rhs);
  Prog& operator=(Prog&& rhs);
  // for normal copy
  Prog(const Prog& rhs);
  Prog& operator=(const Prog& rhs);
  Prog(const ListStatement& p1);
  ~Prog();
  virtual void accept(Visitor *v);
  std::unique_ptr<Program> clone() const override;
};
```

- [WIP] Bison/Flex code genrator switching
  - Still work in progress, but code might require [Bison3.2](https://lwn.net/Articles/770138/)
  - I think I will modify process switching some code in bison

- Use smart pointer in test C++ source
  - This enables test.C to use unique_ptr, however this is just a example. I think this kind of smart-pointer can apply for other C++ code generated by bnfc also. To use smart pointer, I added `#include <memory>`.
  - will generate following code
```cpp
  if (parse_tree)
  {
    printf("\nParse Successful!\n");
    if (!quiet) {
      printf("\n[Abstract Syntax]\n");
      std::unique_ptr<ShowAbsyn> s(new ShowAbsyn());
      printf("%s\n\n", s->show(parse_tree));
      printf("[Linearized Tree]\n");
      std::unique_ptr<PrintAbsyn> p(new PrintAbsyn());
      printf("%s\n\n", p->print(parse_tree));
    }
    delete(parse_tree);
    return 0;
  }
```


hangingman:
@andreasabel Hello, 
I have created PR for using `std::unique_ptr` on test.C as a first step.

hangingman:
@andreasabel 

> Thus, I would like to preserve the old behavior under a flag. 

OK. 
However, how to switch new behavior and old behavior ? 

>Do you think `--ansi` would be a good flag for the C++ backend to preserve the old output?

Yes.
According to the [link of stackoverflow](https://stackoverflow.com/questions/17480543/why-do-ansi-and-std-c11-conflict-in-g), `-ansi` flag represents `-std=c++98` (ISO 1998 C++ standard) in C++ mode.

I could say it's obvious there is big difference between before c++11  and after c++11. So, if we need to preserve the old output, we can select `-ansi (-std=c++98)` or `-std=c++03 (before c++11)`. I think `-ansi` is sufficient. Because it's working so far, and there are few people want to use `c++03`.

hangingman:
@andreasabel OKay, I got it. I'll take a look it

hangingman:
@andreasabel 

>So, instead of making a clone, the code in `CFtoSTLAbs` should be parametrized so that it does slightly different things, depending on whether flag `--ansi` was given or not.

I understand. But, I think it will be very different "ansi C++ code" and "beyond ansi C++ code". So I need to modify functions in `CFtoSTLAbs.hs` different code output, which is concretely `prAbs`, `prCon`, `prList`...

If we introduced flags to switch "ansi" and "not ansi" for each functions, it would be not easy to understand thing.
So, I decided purposely divide `CFtoSTLAbs` modules as `CFtoSTLAbsAnsi` and `CFtoSTLAbsBeyondAnsi`. It's easy to understand and maybe maintainable (?). Haskell has module system it's equivalent to C++ namespace I think.

How do you think ? If you think it should be parametrized, please tell me where we should switch "ansi" and "not ansi" concretely.

andreasabel:
@hangingman Thanks for the reply.  I will have a closer look at the overlap and then answer you.

hangingman:
@andreasabel OK, I guess I know what you are trying to say. https://github.com/BNFC/bnfc/pull/389#issuecomment-948575244
I'll take a look CFtoBisonC.hs and try to refactor `CFtoCPPAbs.hs`.

andreasabel:
@hangingman CI is passing.  Let me know when the PR is ready again for review and testing.

hangingman:
@andreasabel Thank you! 
Now I have added code related with AST, bison/flex part. 
It will take some time to fix rest of them (pritty-print and some). 
I will notify after that, please be patient and wait for a little while 👍

andreasabel:
@hangingman : Happy new year 2022!  How is the progress?

hangingman:
@andreasabel 
Unfortunately my progress is not good. I don't have much time.
I need to modify some code , it was unexpected.
GitHub action is working on my forked repository 
https://github.com/hangingman/bnfc/pull/2

So I will resend pull-request after fixed code.
Thanks!

hangingman:
I close this PR to organize.

PR #391, opened 2021-10-12
C and CPP backends: use size_t for things such as sizes.

enedil:
Previously used `int` is not the return type of functions such as
`strlen`, which generates compiler warnings.


enedil:
I need to note that this probably didn't have any actual correctness impact, as it would require source files greater than 2**31 bytes.

enedil:
Ugh. On my grammar it didn't generate code with such warnings. 

Perhaps https://gcc.gnu.org/onlinedocs/gcc/Diagnostic-Pragmas.html ?

We could push pragma to ignore these kind of errors when including files from flex?

andreasabel:
> I could add a bunch of those, but to be fair, the generated C++ code needs solid rework (as signaled by #293). 

Reworking the C++ backend is most welcome.  Atm, it has no active developer.  (I am just fixing bugs.)

> Parts of the code already assume that std::string is available. Why `PrintAbsyn` does memory allocation by hand to maintain something which is essentially a string?

Not sure.  Maybe because the code for the printer generator is shared between the C++(STL) and C++/NoSTL backends.  

Maybe we could retire the NoSTL backend.  Or keep it while separating it from the STL backend, which then could freely use the STL.  
Since I am not a C++ programmer, I have no feeling for what is standard in the community.  E.g. is there an interest to use C++ without the STL? (Cf. https://softwareengineering.stackexchange.com/questions/161059/is-it-practical-to-abandon-stl-in-c-development)

> What is the stance on slight change of interface? For instance, the PrintAbsyn returns a `char*` while it could be returning an std::string. 

I suppose `std::string` would be in STL only?

I don't mind a modernization of the C++ backends (or just the STL one).  Would you be interested in one and willing to work on it on a larger scale?

enedil:
I'm sorry I didn't have time to answer. Yes, you're correct that this kind of additional flags makes more sense. Does `-Wno-sign-conversion` work with MSVC though?

andreasabel:
> Does `-Wno-sign-conversion` work with MSVC though?

Good question.  I don't know and cannot test this.  In fact, I don't know if any of the `-W` switches works for MSVC...

enedil:
According to https://godbolt.org/z/M1MTaqheE, `-Wall` works, but `-Wno-sign-conversion` doesn't.

enedil:
I agree

PR #399, opened 2021-11-01
Make `closingToken` function pattern matching exhaustive

VBeatrice:
The `closingToken` function cases in the layout solver were not exhaustive

### DIFF
--- a/source/src/BNFC/Backend/Haskell/CFtoLayout.hs
+++ b/source/src/BNFC/Backend/Haskell/CFtoLayout.hs
@@ -206,6 +206,7 @@ cf2Layout layName lexName cf = unlines $ concat
     , "  closingToken :: Position -> Block -> Token"
     , "  closingToken pos = sToken pos . \\case"
     , "    Implicit (LayoutDelimiters _ _ (Just sy)) _ _ -> sy"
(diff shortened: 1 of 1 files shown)


andreasabel:
Mmh, I am not sure.  This gives a better error message, but still `closingToken` will make the layouter crash when it shouldn't, won't it?
What about the round-trip example (parse/print/parse) for the `Alfa.cf` grammar?

VBeatrice:
`closingToken` crashes if one tries to resolve the layouts of a file where the layouts have already been resolved a first time. (when the grammar uses the top level layout pragma)

In the round-trip test the layouts are resolved once, before the first parsing.

andreasabel:
Now I finally found a small reproducer:
```
terminator Exp ";"                         ;
layout toplevel                            ;

Var. Exp ::= Ident;
Let. Exp ::= "let" "{" [Exp] "}" "in" Exp  ;

layout "in"                                ;
layout stop "let"                          ;
```
Input: `let {} in x`

Error is:
```
Layout error at line 1, column 5: trying to close a top level block.
Remaining tokens: { } in x
```

This seems to be a problem with the implementation of `layout stop`.

PR #419, opened 2022-03-31
Haskell: emit `{-# LANGUAGE Safe #-}` (and `Trustworthy` in Lex/Par.hs)

andreasabel:
Haskell: emit `{-# LANGUAGE Safe #-}` (and `Trustworthy` in generated `Lex`/`Par.hs`)
Had to get rid of `GeneralizedNewtypeDeriving` extension in generated `Abs.hs`.

### DIFF
--- a/source/CHANGELOG.md
+++ b/source/CHANGELOG.md
@@ -1,6 +1,7 @@
 # 2.9.5
 
 Unreleased
+* Haskell: label produced code as `Safe` (and parser as `Trustworthy`)
 
 # 2.9.4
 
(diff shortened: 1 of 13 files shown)


andreasabel:
Errors:
```
  * Regression tests:#170 Module Xml cannot be compiled with GADT backend (--xmlt)
  * Regression tests:#170 Module Xml cannot be compiled with GADT backend (--xml)
  * Regression tests:#30 With -d option XML module is not generated inside the directory
  * Parameterized tests:Haskell & Agda:Haskell & Agda:204_InternalToken
  * Parameterized tests:Haskell & Agda (with --functor):Haskell & Agda (with --functor):204_InternalToken
```

andreasabel:
Safe Haskell is stagnant at best, so it does not make sense to accommodate for it.

Issue #423, opened 2022-07-15
Haskell: errors as record types


vkpgwt:
Hello!

BNFC errors in the Haskell backend are plain `String` values like this: ```syntax error at line 3, column 1 before `token'``` . There are some inconveniences here:

- a String is not flexible and customizable enough, as compared with an ADT record type. Currently there is no possibility to get the line and column numbers other than parsing the error message, which is risky in case of future changes in the error message format.
- more information might be provided for better error messages in the ADT record fields: the list of expected token names and maybe the name of the rejected token rule. BNFC Haskell backend is implemented via Happy, which can provide this information.

The new error type might be like:

```hs
data BNFCError = BNFCError
  { bnfcLineNo :: Int
  , bnfcColumnNo :: Int
  , bnfcBadInput :: String
  , bnfcExpectedTokenNames :: [String]
  , bnfcAcceptedTokenName :: String
  }
```

In fact, it might be more complex, since BNFC can also generate lexer errors, and we might want to support the `Text` type.

Since this is a breaking change, a new command line option might be introduced, like `--error-type=(string|record)`, defaulting to the `String` errors.

Would you agree with this change? I would appreciate if you give some advice. I'm using BNFC in production and we need good error messages. It will be perfect if I create a pull request that will be accepted finally, otherwise I'm going to patch BNFC locally.

dpwiz:
> `--errors=string|text|structured`

I'm confused. Having `string|text` selector in multiple places (`--text-token`, maybe there are others) would give spurious combinations like Text tokens / String errors, String tokens / Text errors.

And if the token name type is driven by the existing option, this option would look like `--errors=plain|structured` which is basically a `--structured-errors` flag.

PR #424, opened 2022-08-08
Structured errors in Haskell

vkpgwt:
Hello! This is a proposed implementation of structured errors in Haskell (discussed in #423). Currently it's mostly done, but I guess I need to add some tests, and some issues should be discussed.

- a new option `--errors=string|structured`. It's implemented for Haskell backend only, but its name is abstract enough to be used in other backends in future.
- `Par.y` now exports `Failure` type and some more types to represent parser errors.
- `Par.y` now creates a `Failure` value and converts it to the old-style error string, if string errors are used. Otherwise it returns `Either Failure a` instead of `Either String a`. As I checked manually, the string error format hasn't been changed (maybe we need a test for it?).
- `ErrM.hs` is not created with `--errors=structured`, since I couldn't find a way to modify its content appropriately. And it seems to be for keeping backward compatibility only. Now all error types are exported from `Par.y`.
- I added record syntax to `Posn` type in `Lex.x`, since `Posn` is now used to report error position in exported failure types, and users might deal with it. I find it useful to export `Posn` entirely (absolute position too, not just the line and column numbers). In my project I need to get some context around the error position, which is more convenient to do when we have the absolute position too.

Some issues that might be fixed if needed:

- [ ] no tests added (but a little test to see if `ErrM.hs` is omitted). I guess we need some success tests with `--errors=structured`, and the generated parser tests should be run too - what I have to manually. Could you please make a hint what tests and where I should add?
- [ ] I could not check the new option together with `--glr`, since `--glr` generates broken code on 2.9.4. This PR should make things even worse, since I added some directives to `Par.y` which change the lexer signature from the parser's perspective. I found some comments in the source code which mention that GLR is obsolete. Should I fix it? 
- [ ] I did some refactoring in `Options.hs` to make it possible to write failable option argument parsers, because we need to report an error if `--errors=` gets an invalid argument. I'm not satisfied with the refactoring and its size. Maybe you'd prefer a no-argument option like `--structured-errors` & `--string-errors`? It looks like a typical approach in BNFC: `--glr`, `--text-tokens`, `--bytestrings` (not `--tokens=string|text|bytestring`).
- [ ] In #423 you proposed also `text` argument for `--errors` option which is not supported in the PR, since I'm not sure I understand what it should mean. Should it just change the error type to `Text`, so that `Err` would be `Either Text`?


source/src/BNFC/Options.hs
```diff
@@ -3,6 +3,7 @@
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE RecordWildCards #-}
 {-# LANGUAGE TupleSections #-}
+{-# LANGUAGE ViewPatterns #-}
```
vkpgwt, line 6: It's unneeded

source/src/BNFC/Backend/Haskell/CFtoHappy.hs
```diff
@@ -66,7 +67,14 @@ header modName absName lexName tokenText eps = unlines $ concat
     , "{-# LANGUAGE PatternSynonyms #-}"
     , ""
     , "module " ++ modName
-    , "  ( happyError"
+    , "  ( Err"
+    , "  , Failure(..)"
```
vkpgwt, line 71: Maybe avoid exporting `Failure` & co if `--errors=string`? It might break the backward compatibility if the user imports `Par.hs` unqualified without import list.

source/src/BNFC/Backend/Haskell/CFtoHappy.hs
```diff
@@ -255,24 +269,88 @@ prRules absM functor = vsep . map prOne
...
-footer :: ModuleName -> TokenText -> Bool -> [Cat] -> CF -> String
-footer absName tokenText functor eps _cf = unlines $ concat
+footer :: ModuleName -> TokenText -> Bool -> [Cat] -> ErrorType -> CF -> String
+footer absName tokenText functor eps errorType _cf = unlines $ concat
   [ [ "{"
     , ""
-    , "type Err = Either String"
+    , "-- | The parser failure type."
+    , "--"
+    , "-- It can contain fields of more specific failure record types, so that they"
```
vkpgwt, line 278: "fields" -> "values"

source/src/BNFC/Options.hs
```diff
@@ -376,6 +387,9 @@ specificOptions =
   , ( Option []    ["generic"] (NoArg (\o -> pure o {generic = True}))
           "Derive Data, Generic, and Typeable instances for AST types"
     , haskellTargets )
+  , ( Option []    ["errors"] (ReqArg parseAndSetErrorType "TYPE")
+          "Set the parser error type. Valid values are `string' (default) and `structured'"
```
vkpgwt, line 391: ``Set the parser error type: `string' (default) or `structured'``

vkpgwt:
Hello @andreasabel! Would you mind to look at the PR?

dpwiz:
@andreasabel Is this blocked on something? Any design decisions are yet to be made?

PR #442, opened 2023-04-07
Simplify definition of internal function `columns`

ivanperez-keera:
The internal function columns is building a list with the first element of each sublist, applying an auxiliary function f to it, and then proceeding recursively with the tails of all lists.

This is the same as transposing the list of lists so that all the first elements of the sublists are grouped together, all the second elements are grouped together, and so on, and then applying f to each resulting sublist.

This commit simplifies the definition of columns by using Data.List.transpose.

### DIFF
--- a/source/src/BNFC/Utils.hs
+++ b/source/src/BNFC/Utils.hs
@@ -32,7 +32,7 @@ import Control.Arrow   ((&&&))
 import Control.DeepSeq (rnf)
 
 import Data.Char
-import Data.List          (intercalate)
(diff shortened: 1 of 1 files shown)


andreasabel (commented):
Thanks, indeed! (Quickcheck confirms this).

I think `columns` can now be inlined.
```haskell
widths = map maximum $ transpose $ map (map length) m
```
Do you want to do this step?

ivanperez-keera:
@andreasabel done!

andreasabel (approved):
Thanks!

Issue #394, opened 2021-10-19
A tree-sitter backend


wenkokke:
If we create a [tree-sitter](https://tree-sitter.github.io/tree-sitter/) backend, we could get basic editor support for all languages using BNFC with almost no work.

Tree-sitter grammars are the defacto way of implementing highlighting in [Atom](https://flight-manual.atom.io/hacking-atom/sections/creating-a-grammar/), and there are packages which use tree-sitter grammars to provide highlighting in [VSCode](https://marketplace.visualstudio.com/items?itemName=evgeniypeshkov.syntax-highlighter), [neovim](https://github.com/nvim-treesitter), and [emacs](https://github.com/emacs-tree-sitter/elisp-tree-sitter). There are also [bindings to use a tree-sitter grammar from Java provided by JetBrains](https://github.com/JetBrains/jsitter), which would help with integration into the Jetbrains editor ecosystem.

There are bindings for tree-sitter are in various languages, including Haskell, JavaScript (both Node.js and Wasm), OCaml, Python, Ruby, and Rust.

Compiling a BNF grammar to tree-sitter should be fairly straightforward, and the only major hurdle I foresee would be to implement support for layouts, which would require some custom C code. For an example of how to implement this, one could look at the grammars for [Agda](https://github.com/tree-sitter/tree-sitter-agda), [Haskell](https://github.com/tree-sitter/tree-sitter-haskell), [Python](https://github.com/tree-sitter/tree-sitter-python), or any other language with layout rules.

wenkokke:
Perhaps @banacorn could offer their advice, since they wrote the parser for Agda?￼￼

andreasabel:
> > I wonder whether this could be factored into BNFC -> (E)BNF -> tree-sitter
> 
> Wouldn’t this lead to problems supporting layouts?

Yes, this wouldn't support layout, so maybe it is not worth looking into it, unless BNFC/-layout -> (E)BNF has its own interest.

chaserhkj:
Hi, I have implemented a preliminary tree-sitter backend in #471. If any of the correspondents on this thread are interested, please help me test it out, thank you!

Issue #193, opened 2016-11-04
[Feature Request] Rust Backend


Centril:
`Rust` is quite a new language... but, it is an awesome language.
Particularly, it features `linear types`, `algebraic data types`, and is in addition a `systems programming language`, so it's a nice replacement for `C/C++` and allows for `pattern matching` instead of using the `visitor pattern`.

There are already a `few parser generators`,
+ https://github.com/nikomatsakis/lalrpop
+ https://github.com/kevinmehall/rust-peg
+ https://github.com/goffrie/plex

As well as `parser combinators`:
+ https://github.com/Geal/nom
+ https://github.com/Marwes/combine
+ https://github.com/m4rw3r/chomp

I'll see myself if I can do some development towards this end...
... but mainly, I'm registering this so we can coordinate =)

shlevy:
Is there documentation anywhere on what is needed to add a new backend? I'm playing around with bnfc now and if it fits our use case we'll need to add one for Rust.

andreasabel:
Hello @shlevy, thanks for your initiative!

Atm, there is no such documentation, but I can summarize the most important steps for a new backend:
1. a new `Target` in `Options.hs`: https://github.com/BNFC/bnfc/blob/157f9eb2b6d79308049f6cee92fbab3dc2e3a6ef/source/src/BNFC/Options.hs#L57-L61
  handled in `Main.hs`: https://github.com/BNFC/bnfc/blob/157f9eb2b6d79308049f6cee92fbab3dc2e3a6ef/source/main/Main.hs#L72
2. the backend under https://github.com/BNFC/bnfc/tree/master/source/src/BNFC/Backend
3. a new parameter to the testsuite: https://github.com/BNFC/bnfc/blob/157f9eb2b6d79308049f6cee92fbab3dc2e3a6ef/testing/src/ParameterizedTests.hs#L326-L336

Concerning the backend itself, have a look at the other backends to see their structure and what files they produce.  

[...]

PR #471, opened 2023-11-26
Preliminary implementation of a tree-sitter backend

chaserhkj:
Generates `grammar.js` file that could be used with [tree-sitter](https://tree-sitter.github.io/tree-sitter/) to generate parsers with many different languages. Most of the functionalities of BNFC should be already implemented and working.

There are two remaining issues/features that are not implemented yet:

1. `tree-sitter` does not distinguish between a lexer and a parser, this created issues in their keyword recognition (documented [here](https://tree-sitter.github.io/tree-sitter/creating-parsers#keywords)). They use a keyword extraction mechanism to list token rules that could accept a keyword explicitly. 
    The only way to do this cleanly from the BNFC side is to run all user-defined token patterns against all keywords in the input grammar to determine which patterns should be listed to do keyword extraction. Unfortunately, there are no regex facilities in BNFC to do this currently and an external regex engine is needed. 
    This branch currently just puts all user-defined token patterns into the keyword extraction list for safety. The performance impact of such practice is unknown. How should BNFC approach this problem should be discussed further.
2. Like many other lexers, the regular expression used by `tree-sitter` does not support general set differences. However, `tree-sitter` supports a mechanism known as [external scanners](https://tree-sitter.github.io/tree-sitter/creating-parsers#external-scanners) that could be used to implement this. This is not implemented yet.

These are documented in the source code using `TODO` comments as well.

Should help #193 and #394

### DIFF
--- a/source/src/BNFC/Backend/TreeSitter/RegToJSReg.hs
+++ b/source/src/BNFC/Backend/TreeSitter/RegToJSReg.hs
@@ -10,7 +10,7 @@
 -}
 {-# LANGUAGE LambdaCase #-}
 
-module BNFC.Backend.TreeSitter.RegToJSReg (printRegJSReg) where
+module BNFC.Backend.TreeSitter.RegToJSReg (printRegJSReg, escapeCharFrom) where
 
 import BNFC.Abs
 
--- a/source/src/BNFC/Backend/TreeSitter/RegToJSReg.hs
+++ b/source/src/BNFC/Backend/TreeSitter/RegToJSReg.hs
@@ -1,0 +1,94 @@
+{-
+    BNF Converter: TreeSitter Grammar Generator
+    Copyright (C) 2004  Author:  Markus Forsberg, Michael Pellauer,
+                                 Bjorn Bringert
+
+    Description   : This module converts BNFC Reg to Javascript regular
+                    expressions that is used in
+
+    Author        : Kangjing Huang (huangkangjing@gmail.com)
+    Created       : 23 Nov, 2023
+
+-}
+{-# LANGUAGE LambdaCase #-}
+
+module BNFC.Backend.TreeSitter.RegToJSReg (printRegJSReg) where
+
+import BNFC.Abs
+
+printRegJSReg :: Reg -> String
+printRegJSReg r = "/" ++ render (prt 0 r) ++ "/"
+
+-- There is no space-based formatting for javascript regex
+-- We just concat everything together
+render :: [String] -> String
+render = concat
+
+parenth :: [String] -> [String]
+parenth s = ["("] ++ s ++ [")"]
+
+-- Printer class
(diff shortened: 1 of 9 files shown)


chaserhkj:
Rev: Removed usage of `BlockArguments` to be compatible with older GHC

wenkokke:
I think the primary thing that appears to be missing is support for layouting, for which we'd have to generate custom scanners.

chaserhkj:
Yes, layouting features would require a custom scanner as well. I think that would be an easier starting point for introducing custom scanners for this backend. I would look into it in my spare time.

That being said, the current state of the branch already satisfies my current needs thus it would probably be a lower priority item on my list.

source/src/BNFC/Backend/TreeSitter/RegToJSReg.hs
```diff
@@ -0,0 +1,94 @@
+{-
+    BNF Converter: TreeSitter Grammar Generator
+    Copyright (C) 2004  Author:  Markus Forsberg, Michael Pellauer,
+                                 Bjorn Bringert
```
andreasabel, line 4: I guess you want that copyright notice gone...

andreasabel (commented):
Thanks for the contribution!

There is end-to-end testing for the other backends in the `testing` project (unfortunately not run by CI), how would you go about integrating the tree-sitter backend there?

The matrix for these system tests is set up here: https://github.com/BNFC/bnfc/blob/4272412ea52ba1ff36ee61e4e6e13a7d5da38755/testing/src/ParameterizedTests.hs#L348-L359

source/src/BNFC/Backend/TreeSitter/RegToJSReg.hs
chaserhkj: Done

chaserhkj:
I just added the parameters for testing tree-sitter backend to the testing parameter matrix.

Currently, the backend is only generating `grammar.js` and no test code is generated together with the grammar. So in testing I just called `tree-sitter parse` to use the grammar file to parse input directly.

Unfortunately, there are a number of test failures. I will use next week to try to fix these failures. Before that, I will mark this PR as draft.

chaserhkj:
I am closing this PR since the main commits of it has already been merged (Ref: f1c510157dda55c0f51ca60d04102d3e4688be2a)

A few issues are emerging and related larger changes are made to the tree-sitter backend when I am working on the tests. I will include them in a separate PR.

andreasabel:
> I am closing this PR since the main commits of it has already been merged (Ref: [f1c5101](https://github.com/BNFC/bnfc/commit/f1c510157dda55c0f51ca60d04102d3e4688be2a))

I have not recollection of merging the branch but maybe I did it by accident when I checkout out your PR. :$

PR #483, opened 2024-07-19
Add nix packaging support using haskell.nix

chaserhkj:
This PR adds nix packaging for BNFC using [haskell.nix](https://github.com/input-output-hk/haskell.nix), allowing it to be integrated directly with the nix ecosystem, producing more reproducible builds as well.

Note that this does **not** invalidate the current stack-based build system, this is just additional nix build scripts that can be used by nix systems, quite like the `Dockerfile` we are already providing. This does not create multiple sources of truth for building as well since haskell.nix reads stack yaml configurations and generate build instructions on the fly.


andreasabel:
Thanks for the contribution!

Would be great if this could be tested by a GitHub workflow (CI), so that it does not bit-rot.
Do you think you can add such a workflow?
I have seen it e.g. at https://github.com/haskell/hackage-server/blob/8a8d7391b819921f67aa446c134a2bf83da1b9b4/.github/workflows/nix-flake.yml

chaserhkj:
Workflow is added, should be working now for both os x and linux builds using nix. A few notes:

- Currently I only added building with GHC 9.8 branch in nix workflow. Let me know if you feel building all supported versions makes more sense.
- I saw that they were testing development environments in other example workflows. But our configured development environments includes Haskell language server that needs to be compiled. And that would take a ridiculously longer period of time (>20mins), so I just disabled it for now.
- For `nix flake check` to pass I need to do some workarounds in `flake.nix` to deal with `hydraJobs` and `doctests`, see comments there.

andreasabel:
Excellent, I'll merge this now!

I cannot promise to maintain this, unfortunately, as I lack the `nix` skills.  So further involvement would be appreciated.

PR #484, opened 2024-07-30
Add `parseRawCF` to use bnfc as library and parse LBNF from another application

chaserhkj:
This PR decouples the `CF` parsing process of LBNF file from unused back end options and checks, thus providing a new function `parseRawCF` that can be called from third party applications.

I wrote this mainly because I am building another project and wishes to use LBNF format as my project's input. I want to call BNFC parsing processes directly to parse LBNF in my application but found out that was impossible due to `parseCF` requiring `SharedOptions`. My project performs analysis and manipulation of the input grammar, rather than converting it to an output target format, thus will not use any part of the back end of BNFC.

Ideally we would want to separate `ShareOptions` to front end and back end options and just use front end options for parsing the raw CF. But for now it seems that the parsing process only uses `lbnfFile` option and so I separated this specific field and created the `parseRawCF` function.

Also, I will be implementing a few utilities to work on a parsed LBNF grammar in my project, most significant ones should be the formatter to print `CF` back to LBNF format, and the Chomsky normalization to convert any `CF` to Chomsky Normal Form. I am not sure if these are appropriate to be merged into BNFC as well, let me know you guys' thoughts on them, thank you!

### DIFF
--- a/source/src/BNFC/GetCF.hs
+++ b/source/src/BNFC/GetCF.hs
@@ -18,7 +18,7 @@
 -- | Check LBNF input file and turn it into the 'CF' internal representation.
 
 module BNFC.GetCF
-  ( parseCF
+  ( parseCF, parseRawCF
   , checkRule, transItem
   ) where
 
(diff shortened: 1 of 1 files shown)


andreasabel:
Thanks for the PR and apologies for the long wait.
I have not a lot of time for BNFC as I am mainly developing Agda.

PR #513, opened 2025-08-30
fix lex_destroy return type

ihor-rud:
I noticed that return type of `yylex_destroy` is wrong in `Parser.c`. It was causing weird linking problems for me. This mr fixes it


andreasabel:
Thanks!

PR #563, opened 2026-05-08
[ ci ]: regenerate haskell-ci.yml to fix doctest on GHC >=9.6

Commelina:


Commelina:
Seems there is some problem with doctest-0.25.0 on GHC>=9.6. 

Restrict doctest <0.25 instead of regenerating the CI workflow file.

PR #564, opened 2026-05-08
[ ci ]: restrict doctest<0.25 to avoid error on ghc>=9.4

Commelina:


Commelina:
Maybe use a newer cabal-install is a better choice. See https://github.com/BNFC/bnfc/pull/565 .

